PR #10, opened 2022-02-10
Rasterization with single value for all polygons

zigaLuksic:
This flattens the config (so only one feature is created in a run) and adds the option to specify a value which is used for all polygons. Also removed hardcoded `MASK_TIMELESS` to allow rasterization of `DATA_TIMELESS` features.


eogrow/pipelines/rasterize.py
```diff
@@ -70,11 +71,25 @@ class Schema(Pipeline.Schema):
...
-        columns: List[VectorColumnSchema]
+        output_feature: Feature = Field(description="Output feature of the rasterization.")
+
+        values_column: Optional[str] = Field(
+            description=(
+                "GeoPandas dataframe column name from which to read values for geometries. Use either this or `value`."
+            )
+        )
+        value: Optional[float] = Field(
+            description="Value to use for all rasterized polygons. Use either this or `column_name`."
```
AleksMat, line 82: It should be: "Use either this or `values_column`"

AleksMat (commented):
Overall, I'm not sure if this PR is a step into the right direction. It seems that we might lose more by not being to rasterize the same data in multiple ways than gain by another "simplification" of the pipeline.

zigaLuksic:
> Overall, I'm not sure if this PR is a step into the right direction. It seems that we might lose more by not being to rasterize the same data in multiple ways than gain by another "simplification" of the pipeline.

I agree, in our discussion beforehand this seemed like a good option, but now that the code is here it seems lackluster. I'll close this pull request and make a new one with the other option.

PR #11, opened 2022-02-10
Improve rasterization pipeline

zigaLuksic:
The other option we discussed. Parameters were moved to `columns`, however I am no longer 100% I like the name. Suggestions welcome.


eogrow/pipelines/rasterize.py
```diff
@@ -40,9 +40,29 @@
...
+    )
+    value: Optional[float] = Field(
+        description="Value to use for all rasterized polygons. Use either this or `values_column`."
+    )
+    output_feature: Feature = Field(description="Output feature of rasterization.")
+    polygon_buffer: float = Field(0, description="The size of polygon buffering to be applied before rasterization.")
+    resolution: float = Field(description="Rendering resolution in meters.")
+    overlap_value: Optional[int] = Field(description="Value to write over the areas where polygons overlap.")
+    dtype: str = Field("int32", description="Numpy dtype of the output feature.")
+    no_data_value: int = Field(0, description="The no_data_value argument to be passed to VectorToRasterTaskTask")
```
AleksMat, line 56: Typo "VectorToRasterTaskTask"

eogrow/pipelines/rasterize.py
```diff
@@ -40,9 +40,29 @@
...
+
+    @validator("value")
+    def check_value_settings(cls, v, values):  # pylint ignore:invalid-name,no-self-use,no-self-argument
+        """Ensures that precisely one of `value` and `values_column` is set."""
+        if values["values_column"] is None:
+            assert v is not None, "Only one of `values_column` and `value` should be given."
+        else:
+            assert v is None, "Only one of `values_column` and `value` should be given."
```
AleksMat, line 64: If the error message is the same in both cases then you can simply do

```Python
assert (v is None) != (values["values_column"] is None), "Only one of `values_column` and `value` should be given."
```

PR #14, opened 2022-02-17
Import sort and fix tests

zigaLuksic:
Does two main things:

- fixes tests after new shapely release

- sort imports using `isort`, adds settings to `pyproject.toml` and adds a github action for checking compliance


batic (approved):
Looks good.

PR #19, opened 2022-02-22
BatchDownloadPipeline improvements

AleksMat:


AleksMat:
Improvements of `BatchDownloadPipeline` are finally done:

- `analysis_only` is fully supported,
- batch pipeline has been significantly restructured into more stages,
- grid caching now happens immediately after analysis is done, it should also be less hacky,
- by adding an optional `batch_id` parameter to `BatchAreaManager` we ensure that `BatchAreaManager` can be used independently from `BatchDownloadPipeline`.

Note that changes in this PR depend on PR https://github.com/sentinel-hub/sentinelhub-py/pull/255

eogrow/pipelines/download_batch.py
```diff
@@ -153,10 +187,56 @@ def _create_new_batch_request(self) -> BatchRequest:
...
+            self.batch_client.restart_job(batch_request)
+            LOGGER.info("Restarted partially failed batch job.")
+            return BatchUserAction.START
+
+        LOGGER.info("Didn't trigger batch job because current batch request status is %s", batch_request.status.value)
+        return BatchUserAction.NONE
+
+    def ensure_batch_grid(self, request_id: str) -> None:
+        """This method ensures that area manager caches batch grid into the storage."""
+        self.area_manager = cast(BatchAreaManager, self.area_manager)
```
zigaLuksic, line 228: Just wondering if we would perhaps even REQUIRE that the area manager is  a `BatchAreaManager` or a sublcass of it? :thinking:

zigaLuksic (approved):
Perfect :smile:

Although it does make me a bit sad that it will only work with the `develop` branch of `sh-py`. How do we inform the users? Or do we release a minor version just before `eo-grow` get released?

eogrow/pipelines/download_batch.py
AleksMat: It kinda is required because we rely on the fact that area manager has attribute `batch_id`. Technically, one could write an independent area manager with this attribute and it could still work. But in such case I would recommend to just subclass `BatchAreaManager`.

So for now, I just added casting to make mypy happy. But if you think, I can also add a check in init that makes sure that `self.area_manager` is an instance of `BatchAreaManager`.

AleksMat:
> Although it does make me a bit sad that it will only work with the `develop` branch of `sh-py`. How do we inform the users? Or do we release a minor version just before `eo-grow` get released?

Yes, we will release `sentinelhub-py` `3.5.1` very soon or at least before releasing `eo-grow` `1.1.0`.  And for users that use `eo-grow` from develop branch the standard advice is to also use `sentinelhub-py` and `eo-learn` from develop branch. Our entire stack is being developed in parallel.

eogrow/pipelines/download_batch.py
zigaLuksic: Since we're relying on that, I'd say that maybe we really should check that somewhere in the `__init__`.

AleksMat: Done.

PR #21, opened 2022-02-25
 Add dtype functionality to batch to eopatch conversion

mlubej:


eogrow/pipelines/batch_to_eopatch.py
```diff
@@ -36,6 +36,13 @@ class FeatureMappingSchema(BaseSchema):
     feature_type: FeatureType
     feature_name: str
     multiply_factor: Optional[float] = Field(description="Factor used to multiply feature values with.")
+    dtype: Optional[str] = Field(
+        "float32",
```
mlubej, line 40: @zigaLuksic is this OK?

zigaLuksic: I think the use of a `float32` default is ok here since integer-valued features usually wont be multiplied. I'd just perhaps add a warning to `multiply_factor` that it enables the `dtype` parameter which casts by default. I'm not 100% sure about that yet

AleksMat: Could perhaps the default value here be "auto", meaning that by default no type casting will be done?

zigaLuksic: That happens if `None` is used. Maybe that is the safer default after all, less complicated for users

mlubej: set to `None`.

PR #23, opened 2022-02-28
Improve memory usage in features pipeline

zigaLuksic:
The pipeline now has an option for specifying the output dtype to allow reducing of memory needed for the data.

Despite the recent improvements to `MergeEOPatchTask` it causes some noticeable memory spikes due to data copying. For this reason parts of the pipeline have been linearized by hand (since it does not change the functionality) to deliver a less volatile memory consumption.

### DIFF
--- a/eogrow/pipelines/features.py
+++ b/eogrow/pipelines/features.py
@@ -12,7 +12,6 @@ from eolearn.core import (
     EOWorkflow,
     FeatureType,
     LoadTask,
-    MergeEOPatchesTask,
     MergeFeatureTask,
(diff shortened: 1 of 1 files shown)


zigaLuksic:
The tests fail because setting the dtype relies on functionality that has been recently added to the eo-learn develop :cry:

batic:
> The tests fail because setting the dtype relies on functionality that has been recently added to the eo-learn develop 😢

As that seems to be the only problem, I say it can be merged. Looking at the code everything seems ok.

PR #24, opened 2022-03-02
Add dtype to prediction pipeline

zigaLuksic:
Adds a dtype parameter to prediction pipelines. Found some bugs here and there.

Also updates the tests to check for dtype correctness.


eogrow/pipelines/prediction.py
```diff
@@ -137,6 +140,7 @@ def _get_prediction_node(self, previous_node: EONode) -> EONode:
             input_features=self.config.input_features,
             mask_feature=_optional_typed_feature(FeatureType.MASK_TIMELESS, self.config.prediction_mask_feature_name),
             output_feature=(FeatureType.DATA_TIMELESS, self.config.output_feature_name),
+            output_dtype=self.config.dtype,
```
mlubej, line 143: Perhaps the `output_` prefix is not necessary here

eogrow/tasks/prediction.py
```diff
@@ -81,8 +83,11 @@ def apply_predictor(
             return return_on_empty
 
         if self.mp_lock:
-            return execute_with_mp_lock(predictor, processed_features)
-        return predictor(processed_features)
+            predictions = execute_with_mp_lock(predictor, processed_features)
+        else:
+            predictions = predictor(processed_features)
+
+        return predictions.astype(self.output_dtype) if self.output_dtype else predictions
```
mlubej, line 90: What if the default is that the same `dtype` as input features? Might make more sense.

zigaLuksic: what about classification?

eogrow/pipelines/prediction.py
zigaLuksic: probably not, but i'd still keep it so that it stays more verbose (there are many parameters for the task)

eogrow/tasks/prediction.py
mlubej: Ah, good point... Then I it's most elegant the way it is now.

mlubej: Is there any community preference in such cases between

```python
return predictions.astype(self.output_dtype) if self.output_dtype else predictions
```

and 

```python
if self.output_dtype:
    predictions = predictions.astype(self.output_dtype)

return predictions
```

I can see pros/cons in both, but not sure if this is just personal preference or not.

zigaLuksic: I prefer the first one since it's expression based (and takes up less space), but i'd say it's personal preference

mlubej (approved):

PR #25, opened 2022-03-04
Add test for `BatchToEOPatchPipeline`

zigaLuksic:
As we noticed in the past, this pipeline often goes untested for a while. With this we generate random results of a batch request and then run the pipeline. Suggestions for further tests welcome.


tests/test_pipelines/test_batch_to_eopatch.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+import rasterio
+from fs.base import FS
+
+from eogrow.core.config import Config
+from eogrow.pipelines.batch_to_eopatch import BatchToEOPatchPipeline
+from eogrow.utils.testing import ContentTester, check_pipeline_logs
+
+
+def prepare_batch_files(folder, filesystem: FS, tiff_bbox, width, height, num_timestamps, dtype):
```
AleksMat, line 18: Perhaps you can add types also to the other parameters of this function?

AleksMat (commented):
Just 2 minor things.

PR #32, opened 2022-03-30
Improved "skip existing" functionality and a fix for using it in BatchToEOPatchPipeline.

AleksMat:
Running `BatchToEOPatchPipeline` on 300K patches showed some flaws of "skip existing" functionality. This PR makes basic improvements:

- adds more logs and a progress bar to the filtering procedure,
- renames the filtering function because the previous name `get_patches_without_all_feature` could be interpreted as "get patches that has none of the give features".
- fixes a bug in providing output features in `BatchToEOPatchPipeline` where skipping existing didn't work if EOPatches didn't have timestamps.

A large problem remains - filtering EOPatches is still too slow if we parallelize it only over threads. We could parallelize it also over the cluster. For this I suggest implementing a bit more general parallelization utilities that can be used here. That can be done in future work.

### DIFF
--- a/eogrow/utils/filter.py
+++ b/eogrow/utils/filter.py
@@ -30,7 +30,7 @@ def check_if_features_exist(
     return len(not_seen_features) == 0
 
 
-def get_patches_without_all_features(
+def get_patches_with_missing_features(
     filesystem: FS,
     patches_folder: str,
     patch_list: List[str],
--- a/eogrow/utils/filter.py
+++ b/eogrow/utils/filter.py
@@ -6,6 +6,7 @@ from typing import Iterable, List
 
 import fs
 from fs.base import FS
+from tqdm.auto import tqdm
 
 from eolearn.core.eodata_io import walk_filesystem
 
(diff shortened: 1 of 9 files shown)


eogrow/utils/filter.py
```diff
@@ -30,19 +31,32 @@ def check_if_features_exist(
...
     eopatch_paths = [fs.path.combine(patches_folder, eopatch) for eopatch in patch_list]
 
     def check_patch(eopatch_path: str) -> bool:
         return check_if_features_exist(filesystem, eopatch_path, features)
 
-    with ThreadPoolExecutor() as pool:
-        has_features_iter = pool.map(check_patch, eopatch_paths)
+    with ThreadPoolExecutor() as executor:
+        has_features_list = list(
+            tqdm(
```
zigaLuksic, line 55: is any of our tqdm + threads solutions directly applicable here?

AleksMat: For now I didn't want to complicate with the "proper" solutions as we have in `eo-learn` and `sentinelhub-py` where progress bar correctly measures complete checks with `as_completed`. That is mainly because in this case checking for each EOPatch should take approximately the same amount of time and checking starts from the start of the list.

But in the future I hope that we generalize parallelization utilities (either in `eolearn.core` or `eogrow.utils.parallelize`) where we properly handle progress bar and then we can use those for this use case as well.

zigaLuksic: hmmmm, sure

PR #34, opened 2022-04-05
Update version restrictions

zigaLuksic:
A while ago we had to restrict `responses` since it was breaking our tests. It looks like it can be removed now. We also need to bump the version of `eo-learn` because our pipelines use a feature that was added in the new version.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,4 @@
 black>=22.1.0
-responses
 boto3
 codecov
 flake8
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 black>=22.1.0
-responses < 0.19.0
+responses
 boto3
 codecov
 flake8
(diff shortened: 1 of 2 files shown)


zigaLuksic:
I also restricted the `fs` version as pointed out by @mlubej

requirements-dev.txt
```diff
@@ -1,5 +1,4 @@
 black>=22.1.0
-responses < 0.19.0
```
AleksMat, line 2: Does the issue still exist in version `0.19.0` or was it just not compatible with something else? If there is an issue in `0.19.0` I would suggest setting `responses>=0.20.0`.

AleksMat (approved):

requirements-dev.txt
zigaLuksic: It works with both `0.19.0` and `0.20.0` when I test it locally.

AleksMat: Ok, then it was probably some incompatibility with another package. Then we can leave it without restrictions.

PR #39, opened 2022-04-21
Update config schemas

zigaLuksic:
After the switch to pydantic models and enums in sh-py, a few schemas recieved a small update.

Build work until shpy is released, so this MR can wait a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,6 +102,7 @@ ENV/
 
 # mypy
 .mypy_cache/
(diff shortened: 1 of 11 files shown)


AleksMat (commented):
Looks good. Now that `sentinelhub` `3.5.1.` has been released would you mind setting `sentinelhub>=3.5.1` in requirements? After that feel free to merge.

PR #42, opened 2022-04-25
Grid switching, part 1

AleksMat:
The first part where we make various core changes in order to support grid splitting:

- Extended `AreaManager.cache_grid` and added an abstract method `AreaManager.transform_grid`.
- Extended `BatchAreaManager` to work with `subsplit_x` and `subsplit_y` parameters. Also improved error handling.
- Changed `patch_buffer` parameter in `UtmZoneAreaManager` to parameters `patch_buffer_x` and `patch_buffer_y` which now use work with **absolute buffers**. **[code breaking]**
- Simplified and fixed issues in `EOPatchManager._prepare_names`. Total number of patches is now written in a dataframe by `UtmZoneAreaManager`.
- Added some general utility functions.
- Added unit tests for all this.


eogrow/core/eopatch.py
```diff
@@ -58,28 +59,23 @@ def _prepare_names(self) -> Tuple[bidict, Dict[str, BBox]]:
...
         by a method that generates names in a different way.
 
         :param bbox_dataframe: A dataframe with bounding box geometries and information
-        :param total_patch_num: Total number of EOPatches
         :return: A bidirectional dictionary between names and original IDs
         """
-        if total_patch_num is None:
-            total_patch_num = len(bbox_dataframe.index)
+        if len(bbox_dataframe) == 0:
+            return bidict()
```
zigaLuksic, line 76: is this needed?

eogrow/core/eopatch.py
```diff
@@ -253,6 +249,19 @@ def generate_names(self, bbox_dataframe: GeoDataFrame, *_: Any, **__: Any) -> bi
...
     """A custom patch manager that uses a naming convention based on Sentinel Hub Batch tiles"""
 
-    def generate_names(self, bbox_dataframe: GeoDataFrame, *_: Any, **__: Any) -> bidict:
+    _area_manager: BatchAreaManager
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+
+        if not isinstance(self._area_manager, BatchAreaManager):
+            raise ValueError(f"{self.__class__.__name__} is only compatible with {BatchAreaManager.__name__}")
```
zigaLuksic, line 258: it might be better to update the `Schema` of the manager, so that this is caught at validation time. We do need to test that `pydantic` actually does what it promises (so that it will check that the manager returned by the validator is of the correct type). Keep the `_area_manager: BatchAreaManager` though, it is better than casts as it checks for subtypes correctly

eogrow/pipelines/download_batch.py
```diff
@@ -117,6 +117,13 @@ def __init__(self, *args: Any, **kwargs: Any):
 
         self.batch_client = SentinelHubBatch(config=self.sh_config)
 
+        self.area_manager = cast(BatchAreaManager, self.area_manager)
```
zigaLuksic, line 120: casts can be avoided with the same trick as in the eopatch manager (by annotation)

eogrow/core/eopatch.py
zigaLuksic: This whole fiasco works because under the hood `_area_manager` is of type `Any` (thanks to the `load_manager` function). I am considering adding such a safety net there, somehow (so automatically checking that the annotation and the manager match). Maybe I should open an issue?

eogrow/utils/general.py
```diff
@@ -27,3 +30,38 @@ def jsonify(param: object) -> str:
...
+                "than expected numerical error"
+            )
+
+    return rounded_values.astype(int)
+
+
+def convert_bbox_coords_to_int(bbox: BBox, error: float = 1e-8) -> BBox:
+    """Converts bounding box coordinates to integers by removing numerical errors. If the difference is larger than a
+    numerical error it raises an error."""
+    coords = np.array(list(bbox))
```
zigaLuksic, line 65: does `np.array(bbox)` not work? or perhaps `np.array(*bbox)`?

eogrow/utils/general.py
```diff
@@ -27,3 +30,38 @@ def jsonify(param: object) -> str:
...
+    fixed_coords = convert_to_int(coords, raise_diff=True, error=error)
+    return BBox((*fixed_coords,), crs=bbox.crs)
```
zigaLuksic, line 67: I'd prefer `tuple(fixed_coords)` or maybe `[*fixed_coords]`. Something about `((*fixed_coord,),` absolutely confuses my brain :see_no_evil:

eogrow/core/area/batch.py
```diff
@@ -30,25 +44,35 @@ class Schema(AreaManager.Schema):
...
+        self.absolute_buffer = (
+            self.config.tile_buffer_x * self.config.resolution,
+            self.config.tile_buffer_y * self.config.resolution,
+        )
+
     def _get_grid_filename_params(self) -> List[object]:
-        """Puts together batch parameters"""
-        return [
+        """Puts together batch parameters used for grid filename."""
+        params: List[Any] = [
```
zigaLuksic, line 62: `List[object]` would be a better fit

zigaLuksic (commented):
The split into smaller chunks really helps a lot :)

eogrow/core/eopatch.py
AleksMat: Yes, otherwise `bbox_dataframe.total_num.values[0]` line would fail. In grid switching pipeline this method is being called multiple times on various small dataframes. This handles the case if a dataframe would be empty, although in practice it's not a very likely case.

eogrow/core/eopatch.py
AleksMat: Note that schemas for area manager and EOPatch manager are 2 different subparts of pipeline config. How would you validate that 1 schema is compatible with another? (I wouldn't put this check into the base `Pipeline.Schema` validation because it's way too specific.)

eogrow/pipelines/download_batch.py
AleksMat: You are right. Fixed.

eogrow/utils/general.py
AleksMat: Neither of them works.

eogrow/utils/general.py
AleksMat: Ok, changed to `tuple(fixed_coords)`

eogrow/core/area/batch.py
AleksMat: Fixed.

eogrow/core/eopatch.py
zigaLuksic: ah, nevermind, I got confused on the schema part

Here I would annotate `area_manager` and not `_area_manager`, so that the *type correction* is present in the __init__as well

zigaLuksic: Ah, I see it does not get propagated properly.

zigaLuksic:
Merge when ready

PR #44, opened 2022-04-26
Grid switching, part 3

AleksMat:
Implementation of `SwitchGridsPipeline` pipeline and tasks.

Note: Proper pipeline tests are quite tricky and will be added in the future work.

### DIFF
--- a/eogrow/pipelines/switch_grids.py
+++ b/eogrow/pipelines/switch_grids.py
@@ -32,7 +32,7 @@ class FeatureSchema(BaseSchema):
     )
     unique_columns: Optional[List[str]] = Field(
         description=(
-            "This is only meant to be used for vector features. If provided, it will use the list this columns to drop"
+            "This is only meant to be used for vector features. If provided, it will use this list of columns to drop"
             " duplicated rows in spatially joined dataframes."
         )
     )
@@ -120,12 +120,10 @@ class SwitchGridsPipeline(Pipeline):
         max_input_patch_num = max(len(transformation.source_bboxes) for transformation in transformations)
         load_nodes = [EONode(LoadTask(input_path, config=self.sh_config)) for _ in range(max_input_patch_num)]
 
-        no_data_map = self._get_no_data_map()
-        unique_columns_map = self._get_unique_columns_map()
         join_task = SpatialJoinTask(
             features,
-            no_data_map=no_data_map,
-            unique_columns_map=unique_columns_map,
+            no_data_map=self._get_no_data_map(),
+            unique_columns_map=self._get_unique_columns_map(),
             raise_misaligned=self.config.raise_misaligned,
         )
         join_node = EONode(join_task, inputs=load_nodes)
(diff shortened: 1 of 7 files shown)


eogrow/pipelines/switch_grids.py
```diff
@@ -0,0 +1,237 @@
...
+from ..utils.grid import GridTransformation
+from ..utils.types import Feature, FeatureSpec
+from ..utils.validators import field_validator, optional_field_validator, validate_manager
+
+LOGGER = logging.getLogger(__name__)
+
+
+class FeatureSchema(BaseSchema):
+    feature_type: FeatureType
+    feature_name: Optional[str] = Field(description="A feature name if it exists")
```
zigaLuksic, line 26: why is feature name optional?

eogrow/pipelines/switch_grids.py
```diff
@@ -0,0 +1,237 @@
...
+    no_data_value: float = Field(
+        0,
+        description=(
+            "Used for spatial raster features to fill out parts of a target grid that are outside of a source grid."
+        ),
+    )
+    unique_columns: Optional[List[str]] = Field(
+        description=(
+            "This is only meant to be used for vector features. If provided, it will use the list this columns to drop"
```
zigaLuksic, line 35: typo in description

eogrow/pipelines/switch_grids.py
```diff
@@ -0,0 +1,237 @@
...
+        """Create a workflow that consists of loading, joining groups of EOPatches, slicing into new EOPatches and
+        saving them."""
+        features = self._get_features()
+
+        input_path = self.storage.get_folder(self.config.input_folder_key, full_path=True)
+        max_input_patch_num = max(len(transformation.source_bboxes) for transformation in transformations)
+        load_nodes = [EONode(LoadTask(input_path, config=self.sh_config)) for _ in range(max_input_patch_num)]
+
+        no_data_map = self._get_no_data_map()
+        unique_columns_map = self._get_unique_columns_map()
```
zigaLuksic, line 124: these two could be inlined without information loss

eogrow/tasks/spatial.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def _get_resolution(self, rasters: List[np.ndarray], bboxes: List[BBox]) -> np.ndarray:
+        """Checks that resolutions of given geospatial rasters don't differ more than a rounding error."""
+        resolutions = []
+
+        for raster, bbox in zip(rasters, bboxes):
+            height, width = raster.shape[-3:-1]
+            resolutions.append(np.array(bbox_to_resolution(bbox, width=width, height=height, meters=False)))
+
+        mean_resolution = np.mean(resolutions, axis=0)
```
zigaLuksic, line 68: wouldnt it be better to compare the smallest and largest resolution?

eogrow/tasks/spatial.py
```diff
@@ -0,0 +1,248 @@
...
+        """Provides an empty raster for a given bbox and resolution."""
+        image_size = np.abs(np.array(bbox.upper_right) - np.array(bbox.lower_left)) / resolution
+        try:
+            width, height = convert_to_int(image_size, raise_diff=self.raise_misaligned)
+        except ValueError as exception:
+            raise ValueError(
+                "Pixels from joined raster would be misaligned with pixels from original rasters"
+            ) from exception
+
+        shape = (*sample_raster.shape[:-3], height, width, sample_raster.shape[-1])
```
zigaLuksic, line 88: while I do appreciate how clever this solution is, anyone not familiar with the temporal+nontemporal feature magic would be very confused :)

eogrow/tasks/spatial.py
```diff
@@ -0,0 +1,248 @@
...
+        return raster[..., slice_y, slice_x, :]
+
+    @staticmethod
+    def _filter_vector(gdf: GeoDataFrame, bbox: BBox) -> GeoDataFrame:
+        """Spatial filters a GeoDataFrame."""
+        bbox = bbox.transform(CRS(gdf.crs))
+        intersects_bbox = gdf.geometry.intersects(bbox.geometry)
+        return gdf[intersects_bbox].copy(deep=True)
+
+    def execute(self, eopatch: EOPatch, *, bbox: Optional[BBox]) -> EOPatch:
```
zigaLuksic, line 160: why is the bbox allowed to be empty and what does that signify?

eogrow/tasks/spatial.py
```diff
@@ -0,0 +1,248 @@
...
+    else:
+        raise ValueError("Only one of the resolution and size can be given")
+
+    slice_upper_left = np.array([slice_bbox.min_x, slice_bbox.max_y])
+    slice_lower_right = np.array([slice_bbox.max_x, slice_bbox.min_y])
+    axis_flip = [1, -1]  # image origin is upper left, geographic origin is lower left
+
+    offset = axis_flip * (slice_upper_left - raster_upper_left) / resolution
+    try:
+        offset_x, offset_y = convert_to_int(offset, raise_diff=raise_misaligned)
```
zigaLuksic, line 233: because `convert_to_int` works on `np.ndarrays` the types are set to `Any` after unpacking. Just a warning that annotations should be added if one wants to have more safety when using the function. Also is it ever used on large arrays? so far i've only seen it used on small arrays, and maybe returning `Tuple[int, ...]` would also be ok. Just food for tought

eogrow/pipelines/switch_grids.py
AleksMat: So that also timestamp feature type could be specified here. Otherwise, we would need an additional parameter for timestamp propagation. :thinking:

eogrow/pipelines/switch_grids.py
AleksMat: Fixed.

eogrow/tasks/spatial.py
AleksMat: It's should matter much, besides `mean_resolution` is the thing that will be returned at the end so we are making sure that the values of `mean_resolution` are not problematic.

eogrow/pipelines/switch_grids.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+            save_node = EONode(save_task, inputs=[slice_node])
+            save_nodes.append(save_node)
+
+        return EOWorkflow.from_endnodes(*save_nodes)
+
+    def get_execution_arguments_from_transformations(
+        self, workflow: EOWorkflow, transformations: List[GridTransformation]
+    ) -> List[dict]:
+        """Creates execution arguments."""
```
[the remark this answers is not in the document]
AleksMat: Refactored into multiple methods.

eogrow/tasks/spatial.py
```diff
@@ -0,0 +1,248 @@
...
+        """Spatially slice a raster array."""
+        height, width = raster.shape[-3:-1]
+        slice_x, slice_y = get_array_slices(
+            raster_bbox, slice_bbox, size=(width, height), raise_misaligned=self.raise_misaligned
+        )
+        return raster[..., slice_y, slice_x, :]
+
+    @staticmethod
+    def _filter_vector(gdf: GeoDataFrame, bbox: BBox) -> GeoDataFrame:
+        """Spatial filters a GeoDataFrame."""
```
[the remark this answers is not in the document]
AleksMat: Fixed.

eogrow/tasks/spatial.py
AleksMat: I think it would be weird if `convert_to_int` would besides converting values to ints also convert a numpy array to tuple. Perhaps this issue could be addressed again when we add better numpy types altogether?

eogrow/tasks/spatial.py
AleksMat: Hm, not sure how to make this clearer. I don't like adding if-else statement just for this. Also writing `sample_raster.shape[:-3]` might not be much better.

eogrow/pipelines/switch_grids.py
zigaLuksic: I'm slowly gravitating towards having flags for bbox and timestamp, makes working with other features a breeze. But perhaps outside of scope for this MR. Why not use `FeatureSpec` and have a single param?

eogrow/tasks/spatial.py
zigaLuksic: ok

zigaLuksic:
After you resolve the merge conflicts and decide what to do with the remaining two issues, feel free to merge.

eogrow/tasks/spatial.py
AleksMat: Done.

eogrow/pipelines/switch_grids.py
AleksMat: For now I changed it to a single parameter so that it will be inline with other pipelines.

PR #46, opened 2022-04-29
LocalFolder and improvements in logging and prediction pipelines

AleksMat:
All these changes were done for a use case of having a prediction pipeline that is using a Tensorflow model.

- Parameter `model_filename` was moved from `BasePredictionPipeline` to its 2 child classes. That is because a Tensorflow model consists of a folder structure with multiple files and it is nicer to handle its name a bit differently.
- Implemented `LocalFolder` utility, improved the overall implementation of `LocalFile`/`LocalFolder`, and added more unit tests.
- Improvements in logs:
  * Figured out how to explicitly prevent the crashes because of emitting logs within emitting logs - by adding a default filter `FilesystemFilter` to the `FilesystemHandler`.
  * Fixed minor issues with config parameters.

### DIFF
--- a/eogrow/utils/fs.py
+++ b/eogrow/utils/fs.py
@@ -1,29 +1,33 @@
 """
 Module containing utilities for working with filesystems
 """
+import os
 from typing import Any, Optional
 
 import fs
 import fs.copy
+from fs.base import FS
 from fs.osfs import OSFS
 from fs.tempfs import TempFS
+from fs.walk import Walker
 
 from eolearn.core.utils.fs import get_base_filesystem_and_path
 from sentinelhub import SHConfig
 
 
-class LocalFile:
-    """An abstraction for working with a local version of a file.
+class BaseLocalObject:
+    """An abstraction for working with a local version of remote objects.
 
-    If file's original location is remote (e.g. on S3) then this will ensure working with a local copy. If file's
+    If object's original location is remote (e.g. on S3) then this will ensure working with a local copy. If object's
     original location is already on local filesystem then it will work with that, unless `always_copy=True` is used.
     """
 
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


eogrow/utils/fs.py
```diff
@@ -61,19 +65,19 @@ def __init__(
             self._local_path = fs.path.basename(self._remote_path)
 
         self._absolute_local_path = self._filesystem.getsyspath(self._local_path)
+        self._remote_location_ensured = False
```
zigaLuksic, line 68: what is the usecase where this results in an improvement?

zigaLuksic (approved):

eogrow/utils/fs.py
AleksMat: Something like this:

```Python
with LocalFile("/path/to/file/something.json", mode="w") as local_file:
    # At this point we don't want that path to file is generated yet
    ...
    local_file.copy_to_remote()  # Now the path should be generated
    ...
    local_file.copy_to_remote()  # Path already exists so we don't even have to make IO calls to ensure it exists

...
# Once local_file is closed the file is copied again but again we don't have to ensure that the path exists
```

So basically it aims to reduce a few unnecessary IO calls to remote filesystem.

zigaLuksic: but do we want the user to manually copy the file over multiple times?

AleksMat: In some cases, yes. In our case, we use this for copying logs to a remote filesystem every minute or so. This way, logs are available for users to see right away and if at any time an EC2 instance dies the logs are already on S3.

zigaLuksic: using it in logging is reason enough, good point

PR #47, opened 2022-05-02
Generating test data

AleksMat:
A pipeline and tasks for generating dummy data for testing other pipelines. For now only raster features and timestamps are supported while generating vector features will be a future improvement.

Note that tests fail only because of the issue with `fs`.


eogrow/pipelines/testing.py
```diff
@@ -48,3 +57,82 @@ def run_procedure(self) -> Tuple[List, List]:
...
+                feature_config.feature,
+                shape=feature_config.shape,
+                dtype=np.dtype(feature_config.dtype),
+                min_value=feature_config.min_value,
+                max_value=feature_config.max_value,
+            )
+            node = EONode(task, inputs=[start_node], name=f"{DummyRasterFeatureTask.__name__}_{index}")
+            add_feature_nodes.append(node)
+
+        join_node = EONode(MergeEOPatchesTask(), inputs=add_feature_nodes)
```
zigaLuksic, line 111: what if we want timestamps but no raster features? too much of an edge case?

zigaLuksic: Does `DummyRasterFeatureTask` need the timestamp feature? Otherwise we could have them not depend on it and just join all of them with the merge task?

tests/test_tasks/test_testing.py
```diff
@@ -0,0 +1,63 @@
...
+    assert isinstance(eopatch, EOPatch)
+    assert eopatch.get_feature_list() == [FeatureType.TIMESTAMP]
+    assert len(eopatch.timestamp) == timestamp_num
+    assert eopatch.timestamp == sorted(eopatch.timestamp)
+    assert eopatch.timestamp[0] >= start_time
+    assert eopatch.timestamp[-1] < dt.datetime.fromordinal(end_time.toordinal())
+
+    old_timestamps = eopatch.timestamp
+    eopatch = task.execute(eopatch)
+    assert eopatch.timestamp != old_timestamps
```
zigaLuksic, line 58: you'd probably want to do this with two specific seeds right? otherwise the tests are not deterministic

zigaLuksic (approved):

eogrow/pipelines/testing.py
AleksMat: Good catch - creating EOPatches with no raster features is a completely valid edge case which I missed. I fixed the code now and added an extra test for running a pipeline that generates only timestamps and bboxes.

It turns out that use `MergeEOPatchesTask` if you have time-dependent features but no timestamps. Therefore, timestamps have to be generated before raster features.

tests/test_tasks/test_testing.py
AleksMat: Ah, of course. I changed this test.

PR #48, opened 2022-05-03
Make type of get_execution_arguments more specific

zigaLuksic:


AleksMat (commented):
Looks good. I only see that in `testing.DummyDataPipeline` we use `List[Dict[EONode, Dict[str, Any]]]` while everywhere else its `List[Dict[EONode, Dict[str, object]]]`. Would you mind also updating that? (sorry, I keep mixing `object` and `Any` type :see_no_evil:)

zigaLuksic:
> Looks good. I only see that in `testing.DummyDataPipeline` we use `List[Dict[EONode, Dict[str, Any]]]` while everywhere else its `List[Dict[EONode, Dict[str, object]]]`. Would you mind also updating that? (sorry, I keep mixing `object` and `Any` type see_no_evil)

done.
I guess the best approach is to always go for `object` and only switch to `Any` if there is no other way :thinking: it's really difficult though. `Any` should be used for things that don't have a meaningful type and are used without being inspected (for instance a function loading a ML model). We're bound to mix these two up every now and then :D

PR #49, opened 2022-05-03
Tests for SwitchGridsPipeline

AleksMat:
I also had to:

- extend a functionality in `DummyDataPipeline`,
- fixed an issue with EOPatch order in `SpatialJoinTask`.


eogrow/tasks/spatial.py
```diff
@@ -103,6 +103,10 @@ def execute(self, *eopatches: EOPatch, bbox: BBox) -> EOPatch:
         if any(eopatch.bbox.crs is not bbox.crs for eopatch in eopatches):
             raise ValueError("EOPatches must have the same CRS as the given bounding box")
 
+        # Sorting EOPatches in unique order so that the values in the overlapping areas will always be computed in
+        # the same way.
+        eopatches = tuple(sorted(eopatches, key=lambda eop: repr(eop.bbox)))
```
zigaLuksic, line 108: is the `tuple` required or is it just for safety purposes? (both are valid IMO, just curious)

eogrow/pipelines/testing.py
```diff
@@ -129,16 +149,17 @@ def get_execution_arguments(self, workflow: EOWorkflow) -> List[Dict[EONode, Dic
         """Extends the basic method for adding execution arguments by adding seed arguments a sampling task"""
         exec_args = super().get_execution_arguments(workflow)
 
-        add_feature_nodes = [
-            node
-            for node in workflow.get_nodes()
-            if isinstance(node.task, (DummyRasterFeatureTask, DummyTimestampFeatureTask))
-        ]
+        add_feature_nodes = list(self._nodes_to_configs_map)
         add_feature_nodes.sort(key=lambda node: node.get_name())  # To ensure seeds are always given in the same order
```
zigaLuksic, line 153: would it work in one line if you used `sorted` here?

AleksMat: Good point, I updated. I also renamed `node` -> `_node` because PyCharm was telling me `node` variable is already used in the outer scope.

PR #50, opened 2022-05-04
specialize all type: ignore comments and get rid of some

zigaLuksic:
When updating another project I remembered that `ŧype: ignore` has an option to specify precisely the type of error you're ignoring. Since this is much better than just ignoring all errors in a line, I re-did all such comments.

In some cases I also manged to avoid the issue by adjusting something else, reducing the number of ignore comments from 20 to 14 :tada:


zigaLuksic:
Still need to fix some issues with the manager validation :(

AleksMat (approved):

PR #53, opened 2022-05-16
Allows rasterization of temporal/timeless vector features.

batic:
Minimal changes that allow rasterizing temporal vector features. If this is merged, then the pipeline can be extended with only the `build_pipeline` method overridden in order to first load existing patches (to have timestamps), and then temporarily add (temporal) vector features that are rasterised.

### DIFF
--- a/eogrow/pipelines/rasterize.py
+++ b/eogrow/pipelines/rasterize.py
@@ -100,14 +100,14 @@ class RasterizePipeline(Pipeline):
         super().__init__(*args, **kwargs)
 
         self.filename: Optional[str] = None
-        assert len(set([c.output_feature[0].is_temporal() for c in self.config.columns])) == 1, \
-            "All output features have to be either temporal or timeless!"
+
+        if len({c.output_feature[0].is_temporal() for c in self.config.columns}) != 1:
+            raise ValueError("All output features have to be either temporal or timeless!")
 
         if isinstance(self.config.vector_input, str):
             self.filename = self._parse_input_file(self.config.vector_input)
-            feature_type = FeatureType.VECTOR_TIMELESS
-            if self.config.columns[0].output_feature[0].is_temporal():
-                feature_type = FeatureType.VECTOR
+            output_is_temporal = self.config.columns[0].is_temporal()
+            feature_type = FeatureType.VECTOR if output_is_temporal else FeatureType.VECTOR_TIMELESS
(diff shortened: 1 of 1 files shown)


eogrow/pipelines/rasterize.py
```diff
@@ -100,9 +100,15 @@ def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
 
         self.filename: Optional[str] = None
+        assert len(set([c.output_feature[0].is_temporal() for c in self.config.columns])) == 1, \
```
zigaLuksic, line 103: use `any` instead of checking length. Also we try to raise `ValueError` instead of asserts (for error catching purposes)

eogrow/pipelines/rasterize.py
```diff
@@ -100,9 +100,15 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+            "All output features have to be either temporal or timeless!"
+
         if isinstance(self.config.vector_input, str):
             self.filename = self._parse_input_file(self.config.vector_input)
-            self.vector_feature = FeatureType.VECTOR_TIMELESS, f"TEMP_{uuid.uuid4().hex}"
+            feature_type = FeatureType.VECTOR_TIMELESS
+            if self.config.columns[0].output_feature[0].is_temporal():
+                feature_type = FeatureType.VECTOR
```
zigaLuksic, line 110: I'd prefer something along the lines of:
```python
output_is_temporal = self.config.columns[0],..
feature_type = FeatureType.VECTOR if output_is_temporal else FeatureType.VECTOR_TIMELESS
 ```

File: eogrow/pipelines/rasterize.py (hunk 1 above)
Comment (AleksMat) [reply eogrow/pipelines/rasterize.py:103]: Also you can remove an unnecessary list there and make a set directly from generator:

```Python
set(c.output_feature[0].is_temporal() for c in self.config.columns)
```

zigaLuksic: as discussed privately `all` and `any` get a bit long because they just all need to be equal.
So i just suggest to use set comprehension, i.e.
```python
 len({c.output_feature[0].is_temporal() for c in self.config.columns}) == 1
 ```

batic: Changed.

eogrow/pipelines/rasterize.py
batic: Changed.

AleksMat (approved):

PR #56, opened 2022-05-17
Ray utils

AleksMat:
I extracted the procedure of connecting to Ray into a utility function. This way I can use it for a FD pipeline parallelization which works outside of the scope of EOExecutor. I also added tests for this utility function.

### DIFF
--- a/eogrow/utils/ray.py
+++ b/eogrow/utils/ray.py
@@ -1,0 +1,35 @@
+"""
+Modules with Ray-related utilities
+"""
+import logging
+from typing import Literal, Union
+
+import ray
+
+LOGGER = logging.getLogger(__name__)
+
+
+def connect_to_ray(use_ray: Union[Literal["auto"], bool] = "auto") -> bool:
+    """According to the given parameter it will try to connect to an existing Ray cluster.
+
+    :param use_ray: Either a boolean flag or `"auto"` to define if the connection should be established or not.
+    :return: `True` if connection is established and `False` otherwise.
+    """
+    if use_ray == "auto":
+        try:
+            _try_connect_to_ray()
+            return True
+        except ConnectionError:
+            LOGGER.info("No Ray cluster found, will not use Ray.")
+            return False
+
+    if use_ray:
+        _try_connect_to_ray()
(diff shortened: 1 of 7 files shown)


eogrow/utils/types.py
```diff
@@ -12,3 +12,5 @@
 
 Feature = Tuple[FeatureType, str]
 FeatureSpec = Union[Tuple[FeatureType, str], FeatureType]
+
+BoolOrAuto = Union[Literal["auto"], bool]
```
zigaLuksic, line 16: do you suppose such a type will occur often?

eogrow/utils/ray.py
```diff
@@ -0,0 +1,36 @@
...
+import logging
+
+import ray
+
+from .types import BoolOrAuto
+
+LOGGER = logging.getLogger(__name__)
+
+
+def connect_to_ray(use_ray: BoolOrAuto = "auto") -> bool:
```
zigaLuksic, line 13: not sure about the choice of name, especially `connect_to_ray(use_ray=False)` sounds a bit odd. Don't have suggestions so far.

zigaLuksic (approved):

eogrow/utils/types.py
AleksMat: I think the type is general enough that it can occur more often in the future for various pipeline parameters. This way it is also clearer that the type that is used in pipeline schema and in the utility function is the same. Also in general we should try to avoid code duplications of complicated type definitions (so far, I also wasn't paying much attention about that myself).

eogrow/utils/ray.py
AleksMat: Hm, I also can't find a better name for this. :thinking: 

Atm, I can think of `resolve_ray_connection`, `setup_ray`, `ensure_ray_requirement`. :thinking:

AleksMat: Renamed to `handle_ray_connection`. I'm not 100% happy but it's better than before.

zigaLuksic: I also like it better

PR #59, opened 2022-05-31
Better batch bbox reconstruction

zigaLuksic:
This MR switches BBox reconstruction from rounding geometries to recalculating with the given information, as suggested.


batic (approved):
Looks good.

PR #60, opened 2022-05-31
Improve RAM handling of the batch to eopatch pipeline

mlubej:
The pipeline loads all the features and then saves them with a single task. This can blow up the memory and forces the user to work with underutilized CPU.

The changes here follow some recent logic discussed with @zigaLuksic, where each feature is separately loaded, saved and then removed. Loading multiple tiffs into a single feature is still supported, but here all provided features are loaded into memory before saving.

Below is the before/after RAM usage for a single 10km x 10km EOPatch at 10 m res.

image

Left is this MR, right is old behavior.

### DIFF
--- a/eogrow/pipelines/batch_to_eopatch.py
+++ b/eogrow/pipelines/batch_to_eopatch.py
@@ -110,9 +110,9 @@ class BatchToEOPatchPipeline(Pipeline):
 
     def build_workflow(self) -> EOWorkflow:
         """Builds the workflow"""
-        userdata_node = None
+        last_node = None
         if self._has_userdata:
-            userdata_node = EONode(
+            last_node = EONode(
                 LoadUserDataTask(
                     path=self._input_folder,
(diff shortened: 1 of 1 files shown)


eogrow/pipelines/batch_to_eopatch.py
```diff
@@ -110,9 +110,9 @@ def _get_output_features(self) -> List[FeatureSpec]:
 
     def build_workflow(self) -> EOWorkflow:
         """Builds the workflow"""
-        userdata_node = None
+        last_node = None
```
zigaLuksic, line 113: just for the sake of consistency, we tend to use `previous_node` (since `last` slightly implies it is the last node of the workflow)

mlubej: fixed

mlubej:
Just to give further content, I ran this updated pipeline 4 times using all cores, it achieved 100% CPU utilization, RAM was stable all the while it was running. All runs successful.

PR #65, opened 2022-06-03
Session transfer in download pipeline

zigaLuksic:
I am not yet completely happy with how information about the execution kind is obtained/propagated.

Was thinking that maybe it could be a property of the pipeline? Not sure...
The biggest issue is that one must connect to ray just to figure if the cluster exists. However in theory we want to do that just before execution, and not at initialization, since the environment might change (though very unlikely).


eogrow/utils/types.py
```diff
@@ -14,3 +14,5 @@
 FeatureSpec = Union[Tuple[FeatureType, str], FeatureType]
 
 BoolOrAuto = Union[Literal["auto"], bool]
+
+ExecutionKind = Literal["single", "multi", "ray"]
```
AleksMat, line 18: I would definitely make this an `Enum` so that we don't have to work with strings in the code. I see that we already have `eolearn.core.utils.parallelize._ProcessingType` which is basically what we want here. But I'm not sure if we should make that enum public or recreate it here :thinking:.

eogrow/core/pipeline.py
```diff
@@ -156,6 +157,13 @@ def get_execution_arguments(self, workflow: EOWorkflow) -> List[Dict[EONode, Dic
             exec_args.append(single_exec_dict)
         return exec_args
 
+    def _prepare_execution(self) -> ExecutionKind:
```
AleksMat, line 160: As you wrote, this could also be a property. But since it has a side-effect of initializing ray it might be a bit too misleading to have it as a property. Therefore, I'm ok if it remains a method.

Regarding the name alternatives, I was thinking about `init_execution` or even `init_processing` because this isn't limited to "EOExecution" as defined in eo-learn.

eogrow/pipelines/download.py
```diff
@@ -135,6 +170,20 @@ def get_execution_arguments(self, workflow: EOWorkflow) -> List[Dict[EONode, Dic
 
         return exec_args
 
+    def run_procedure(self) -> Tuple[List[str], List[str]]:
+        execution_kind = self._prepare_execution()
+        session_loader = self._create_session_loader(execution_kind)
```
AleksMat, line 175: Because calling `self._prepare_execution` doesn't really cost anything, you could restructure this a bit so that `self._create_session_loader` doesn't have to receive `execution_type` but instead just calls `self._prepare_execution` itself. This means that `self._create_session_loader` could be called only in `_get_download_node` method and not sooner. 

- advantage: you wouldn't have to propagate `session_loader` parameters 2 levels deep,
- disadvantage: `self._prepare_execution` gets called 1 more time.

eogrow/utils/types.py
zigaLuksic: I don't mind keeping it as a string since we're mypy-checked, but I can also switch it to Enum.

eogrow/pipelines/download.py
zigaLuksic: because `_prepare_execution` has side-effects I went with value propagation. Another option is to assign the result to an attribute and read from it in `_get_download_node` but then the execution model becomes more complex since the two methods communicate over non-standard channels

AleksMat: Yes, but the side effect can happen multiple times without problems. Ray starts a connection with cluster only the first time.

Although I see also your point that we might want to be more explicit with this and not to hide the side-effect in some sub-method. So I'm leaving it up to you to decide.

eogrow/utils/types.py
AleksMat: The only extra benefit of having an enum over literal is that you get auto-complete when you write the code. E.g. when I write `if execution_kind == "ray":` I don't get an auto-complete on `"ray"` in PyCharm.

eogrow/pipelines/download.py
zigaLuksic: I was thinking that perhaps each pipeline has an attribute `execution_mode` (by default 'single'), and that is *adjusted* in the `_prepare_execution`. However this has precisely the issue described above, if the adjustment hasn't happened yet the value is wrong. The big issue here is that with 'auto' we don't know if ray is available or not. The price of comfort :)

I would leave it as it is, because it also explicitly states the requirement that the workflow is adjusted according to the execution mode, perhaps a good thing.

eogrow/utils/types.py
zigaLuksic: Since we're using enums extensively throughout our codebase I think it is the correct thing to do so here as well. Will adjust.

eogrow/pipelines/download.py
AleksMat: Ok, let's leave it as it is.

eogrow/utils/types.py
zigaLuksic: done

eogrow/core/pipeline.py
AleksMat: Nice. Now that we switched to name `_init_processing`, should we also rename `ExecutionKind` into `ProcessingKind` or `ProcessingType`?

PR #66, opened 2022-06-06
Batch2eopatch pipeline can adjust dtype without setting multiplication factor

zigaLuksic:
Just a small quality of life improvement


eogrow/pipelines/batch_to_eopatch.py
```diff
@@ -36,11 +36,7 @@ class FeatureMappingSchema(BaseSchema):
     )
     feature: Feature
     multiply_factor: Optional[float] = Field(description="Factor used to multiply feature values with.")
```
AleksMat, line 38: I'm just wondering if perhaps we can already here set:

```python
multiply_factor: float = Field(1, description="Factor used to multiply feature values with.")
```

and then in the pipeline just check:

```
if mapping.multiply_factor != 1 or mapping.dtype is not None:
```

That is because in this case `1` is the default value that will never change a feature. So distinguishing between `None` and `1` seems unnecessary.

AleksMat (commented):
Looks good, only 1 comment.

eogrow/pipelines/batch_to_eopatch.py
zigaLuksic: good idea, changed

AleksMat: Excellent, the only thing left is that you can now remove line:

```Python
factor = mapping.multiply_factor if mapping.multiply_factor is not None else 1
```

PR #73, opened 2022-06-20
Add utility function for running dummy data generating pipelines

zigaLuksic:


eogrow/utils/testing.py
```diff
@@ -266,3 +264,31 @@ def create_folder_dict(config_folder: str, stats_folder: str, subfolder: Optiona
         "config_folder": os.path.join(config_folder, subfolder) if subfolder else config_folder,
         "stats_folder": os.path.join(stats_folder, subfolder) if subfolder else config_folder,
     }
+
+
+def run_dummy_data_pipeline(
```
AleksMat, line 269: I'm not entirely sure we need this utility function. We already have `run_and_test_pipeline` and you can always write a testing procedure in a way that you chain 2 pipeline configs in the same config file and then execute that with `run_and_test_pipeline`. An example of that is in `test_switch_grids.py` pipeline test. So I'm not sure when would that procedure not be enough and this function needed instead. :thinking:

eogrow/utils/testing.py
```diff
@@ -266,3 +264,31 @@ def create_folder_dict(config_folder: str, stats_folder: str, subfolder: Optiona
...
+) -> None:
+    """A default way of testing a pipeline
+
+    :param config_file: Name of the config file containing a dummy data generating pipeline
+    :param config_folder: A path to folder containing the config file
+    :param folder_key: Folder containing results of the pipeline, if missing it's inferred from config
+    :param reset_folder: If True it will delete content of the folder with results before running the pipeline
+    """
+    if not config_file.endswith(".json"):
+        config_file += ".json"
```
AleksMat, line 280: Perhaps we could even raise an error in case the name doesn't end with `".json"`? Automatic adding of file format can produce weird file names.

eogrow/utils/testing.py
zigaLuksic: hmm, chaining configs is indeed a way to skip this :thinking: 
But would the output folder of the dummy pipeline (input folder of the proper pipeline) be *cleaned* ?

AleksMat: Yes, the output of every pipeline in the chain gets removed before that pipeline is run: https://github.com/sentinel-hub/eo-grow/blob/v1.1.1/eogrow/utils/testing.py#L240

The only improvement I could think of here is that we could perhaps first delete contents of all output folders and only then run the pipeline chain. This way we also support pipelines that have same input and output folders.

zigaLuksic: lets close this MR for now and then re-think the whole testing suite at one point

PR #80, opened 2022-07-26
Improve tests w.r.t. mirroring and other accidental reordering

zigaLuksic:
When trying to detect the bug in the batch2eopatch pipeline I discovered that our tests are very poor when it comes to accidental reordering of arrays. The random points have a certain sensitivity, but it's not always clear whether that could have been just accidental due to different rounding errors.

I included a **subsample test** that calculates the base statistics on a random 10% sample of points. Since the points are chosen randomly (with a fixed seed), rounding errors should not impact the result, but any reordering should.

I also switched the random points to using `np.random.Generator`


eogrow/utils/testing.py
```diff
@@ -180,6 +180,11 @@ def _calculate_numpy_stats(self, raster: np.ndarray) -> JsonDict:
                 for name, operation in self._STATS_OPERATIONS.items()
             }
 
+            subsample = self._get_subsample(finite_values)
+            stats["subsample_basic_stats"] = {
+                name: self._prepare_value(operation(subsample)) for name, operation in self._STATS_OPERATIONS.items()
+            }
```
AleksMat, line 186: Maybe it would be a bit nicer if you moved all more code into the the submethod that currently has only `2` lines. You could have something like this:

```Python
stats["subsample_basic_stats"] = self._get_subsample_stats(finite_values)
```

AleksMat (commented):
Looks good, just a single minor comment.

eogrow/utils/testing.py
zigaLuksic: done

PR #81, opened 2022-07-26
Added size parameter to download pipelines and unified multi-parameter validations

AleksMat:
Changes:

- Users can now choose between resolution and size parameter for download.
- Added a custom type `RawPipelineDict` and unified how validators are implemented in pipelines. Whenever we have multi-parameter validation (without parsing) it seems to me that `@root_validator` is always a better option over `@validator("one parameter name")`.


eogrow/utils/types.py
```diff
@@ -17,6 +17,7 @@
 BoolOrAuto = Union[Literal["auto"], bool]
 
 JsonDict = Dict[str, Any]
+RawPipelineDict = Dict[str, Any]
```
zigaLuksic, line 20: i am not entirely happy with this name, but not sure what else it could be
its not RawConfig, because the values have already been parsed...

AleksMat: Maybe `RawSchemaDict`? I can't think of any much better name. :thinking:

zigaLuksic: Lets go with that

AleksMat: Ok, renamed to `RawSchemaDict`.

PR #84, opened 2022-07-28
ExportMapPipeline ignores dtype and nodata when merging

mlubej:
In the ExportMapPipeline the tiff exporting task correctly exports the dtype and the no_data part of the EOPatches, but this information doesn't get passed on when the separate tiffs get merged using `gdal_merge.py` and later on in the cogification part.

This MR:
- passes on the `dtype` and the `nodata` value from the pipeline config to the `_merge_map_function`
- converts the python-based `dtype` string to the GDAL-specific combo (workaround for missing 8-bit tiffs)
- passes the `dtype` and `nodata` wherever they can be accepted (`gdal_merge.py` and `gdal_translate`)
- improves the `gdal_merge.py` subprocess string

### DIFF
--- a/eogrow/utils/map.py
+++ b/eogrow/utils/map.py
@@ -6,16 +6,23 @@ import os
 import shutil
 import subprocess
 from tempfile import NamedTemporaryFile
-from typing import List, Union
+from typing import List, Literal, Optional, Union
 
 LOGGER = logging.getLogger(__name__)
 
 
-def cogify_inplace(tiff_file: str, blocksize: int = 2048, nodata: Union[None, int, float] = None) -> None:
+
+def cogify_inplace(
+    tiff_file: str,
+    blocksize: int = 2048,
+    nodata: Union[None, int, float] = None,
+    dtype: Optional[Literal["int8", "int16", "uint8", "uint16", "float32"]] = None,
+) -> None:
     """Make the (geotiff) file a cog
     :param tiff_file: .tiff file to cogify
     :param blocksize: block size of tiled COG
     :param nodata: value to be treated as nodata, default value is None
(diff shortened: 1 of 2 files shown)


eogrow/utils/map.py
```diff
@@ -6,21 +6,35 @@
...
+    "int16": "-ot Int16",
+    "float32": "-ot Float32",
+}
 
-def cogify_inplace(tiff_file: str, blocksize: int = 2048, nodata: Union[None, int, float] = None) -> None:
+
+def cogify_inplace(
+    tiff_file: str,
+    blocksize: int = 2048,
+    nodata: Union[None, int, float] = None,
```
mlubej, line 25: @zigaLuksic would `Optional[Union[int, float]]` work better here, or is it the same?

eogrow/utils/map.py
```diff
@@ -6,21 +6,35 @@
...
+    dtype: Optional[Literal["int8", "int16", "uint8", "uint16", "float32"]] = None,
```
zigaLuksic, line 26: you can do `Literal[None, "int8", ...]`, less brackets, same in line 46 and 127

mlubej: fixed

eogrow/utils/map.py
mlubej: I guess similarly as in the other comment, less brackets is better

zigaLuksic: In fact here `Optional[float]` would be the correct. When these were typed we didnt yet know the python mypy convention that `Union[int, float]` is annotated as just `float`.

mlubej: I see, will reduce, thanks

mlubej: done

PR #85, opened 2022-07-29
Handling of nonexistent aws_profile

AleksMat:
This PR changes that in case given AWS profile doesn't exist a warning is shown instead of an error. It also adds a test that checks this.

The reason for this is purely practical - if you run a pipeline on AWS instance with a role that already allows accessing S3 buckets you don't need AWS profiles. But if you run a pipeline locally then you need to specify `aws_profile` in order to access S3 buckets. So to avoid constantly changing `aws_profile` parameter it seems easier to just give a warning instead of an error. Although I'm not 100% sure if this really justifies this change. :thinking:

### DIFF
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -1,15 +1,13 @@
 """
 This module handles everything regarding storage of the data
 """
-import warnings
 from io import StringIO
 from typing import Dict, List, Optional
 
 import fs
-from botocore.exceptions import ProfileNotFound
-from pydantic import Field
+from fs.base import FS
+from pydantic import BaseSettings, Field
 
-from eolearn.core.exceptions import EOUserWarning
 from eolearn.core.utils.fs import get_aws_credentials, get_filesystem, is_s3_path
 from sentinelhub import SHConfig
 
@@ -21,7 +19,7 @@ from .schemas import ManagerSchema
 class StorageManager(EOGrowObject):
     PRESET_FOLDERS: Dict[str, str] = {"logs": "logs", "input_data": "input-data", "cache": "cache"}
 
-    class Schema(ManagerSchema):
+    class Schema(ManagerSchema, BaseSettings):
         project_folder: str = Field(
             description=(
                 "The root project folder. Can be either local or on AWS S3 Bucket."
@@ -29,10 +27,11 @@ class StorageManager(EOGrowObject):
             ),
         )
         aws_profile: Optional[str] = Field(
+            env="AWS_PROFILE",
             description=(
-                "The AWS profile with credentials needed to access the S3 buckets. In case the profile doesn't exist it"
-                " will show a warning."
-            )
+                "The AWS profile with credentials needed to access the S3 buckets. In case the profile isn't specified"
+                " with a parameter it can be read from an environmental variable."
+            ),
         )
         aws_acl: Optional[AwsAclType] = Field(
             description=(
@@ -44,6 +43,9 @@ class StorageManager(EOGrowObject):
             default_factory=dict, description="A flat key: value store mapping each key to a path in the project."
         )
 
+        class Config(ManagerSchema.Config):
+            case_sensitive = True
+
     config: Schema
 
     def __init__(self, config: Schema):
@@ -54,12 +56,7 @@ class StorageManager(EOGrowObject):
                 self.config.structure[folder_key] = folder_path
 
         self.sh_config = self._prepare_sh_config()
-
-        fs_kwargs: Dict[str, str] = {}
-        if is_s3_path(self.config.project_folder) and self.config.aws_acl:
-            fs_kwargs["acl"] = self.config.aws_acl
-
-        self.filesystem = get_filesystem(self.config.project_folder, create=True, config=self.sh_config, **fs_kwargs)
+        self.filesystem = self._prepare_filesystem()
 
     def _prepare_sh_config(self) -> SHConfig:
         """Prepares an instance of `SHConfig` containing AWS credentials. In case given AWS profile doesn't exist it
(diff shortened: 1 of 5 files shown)


zigaLuksic (commented):
I am not sure i'd want to introduce a silent failure just because it is more convenient for our specific usecase

AleksMat:
> I am not sure i'd want to introduce a silent failure just because it is more convenient for our specific usecase

I changed the functionality to read from an environmental variable, as we discussed. The new way nicely solves our problems.

Notice that I had to make `StorageManager.Schema` inherit for `pydantic.BaseSettings`. Class `pydantic.BaseSettings` is a subclass of `pydantic.BaseModel` with an additional functionality to handle environmental variables and secrets. Looking at its implementation I don't see anything problematic so now I'm wondering if we should perhaps make even our base `EOGrowObject.Schema` inherit from `BaseSettings` instead of `BaseModel`? This would in future allow us to add more cases of reading from env variables and secrets. Perhaps it's only worth benchmarking if there is any noticeable performance decrease of using `BaseSettings` instead of `BaseModel`.

AleksMat:
I spoke too soon. There is one unexpected extra functionality that comes with `BaseSettings` - every parameter from `BaseSettings` model can be overwritten with an environmental variable with the same name.

E.g. if you don't provide `project_folder` parameter in config file it can be taken from an environmental variable `project_folder`. This can be a bit dangerous because it can lead to some unwanted parameters. However, we can configure this by setting `BaseSettings.Config.env_prefix` to e.g. `"eogrow_"` like I did in this PR. This means that it would instead look for env variable `eogrow_project_folder` which is very unlikely to exist unless user explicitly defines it.

tests/test_core/test_storage.py
```diff
@@ -68,6 +68,36 @@ def test_get_custom_folder(local_storage_manager: StorageManager, project_folder
...
+    also that config parameter takes priority over environmental variable."""
+
+    for parameter_key, parameter_value, config_dict in [
+        ("aws_profile", config_profile, aws_storage_config),
+        ("AWS_PROFILE", env_profile, os.environ),
+    ]:
+        if parameter_value is not None:
+            config_dict[parameter_key] = parameter_value
+        elif parameter_key in config_dict:
+            del config_dict[parameter_key]
```
zigaLuksic, line 84: The setup step is not straightforward to understand (the idea of what we are trying to accomplish here), some more info in the description would be great for maintainability

tests/test_core/test_storage.py
```diff
@@ -68,6 +68,36 @@ def test_get_custom_folder(local_storage_manager: StorageManager, project_folder
...
+
+    try:
+        expected_profile = config_profile if config_profile is not None else env_profile
+        if expected_profile:
+            with pytest.raises(ProfileNotFound) as exception_info:
+                StorageManager.from_raw_config(aws_storage_config)
+
+            assert str(exception_info.value) == f"The config profile ({expected_profile}) could not be found"
```
zigaLuksic, line 92: isn't it odd that we have an `expected_profile` despite expecting the profile not to exist?

tests/test_core/test_storage.py
```diff
@@ -68,6 +68,36 @@ def test_get_custom_folder(local_storage_manager: StorageManager, project_folder
...
+        else:
+            storage = StorageManager.from_raw_config(aws_storage_config)
+            assert storage.config.aws_profile == expected_profile
+    finally:
+        if "AWS_PROFILE" in os.environ:
+            del os.environ["AWS_PROFILE"]
```
zigaLuksic, line 98: wouldn't this mess up the users profile settings?

eogrow/core/storage.py
```diff
@@ -18,15 +19,19 @@
 class StorageManager(EOGrowObject):
     PRESET_FOLDERS: Dict[str, str] = {"logs": "logs", "input_data": "input-data", "cache": "cache"}
 
-    class Schema(ManagerSchema):
+    class Schema(ManagerSchema, BaseSettings):
```
zigaLuksic, line 22: so this inherits from both the `BaseModel` and `BaseSettings`... Does that cause any clashes? Does it still complain if a parameter is not set or an extra parameter is given? (I see that you extended the config, i'm just curious if its done correctly)

zigaLuksic (commented):
I've been hesitant about using BaseSettings because of 'environment absorption', however you handled it nicely with the prefix. I didn't know you could add a prefix and then override that with `env="AWS_PROFILE"`

tests/test_core/test_storage.py
AleksMat: No. At the start of Python process, env variables from external environment are copied to Python process. But when Python process ends, any env variable set or changed by Python process are not applied back to external environment. Hence nothings outside the scope of tests will get messed up. But we have to do this step, otherwise this would mess up other tests.

zigaLuksic: ah, as long as it doesnt mess up the environment outside of python this is fine

eogrow/core/storage.py
AleksMat: As far as I can tell, everything works as it should and the config has been extended correctly. Looking at the implementations of these `pydantic` objects there are no implementation clashes.

As I mentioned, there is also an option that we change `EOGrowObject.Schema` to already inherit from `BaseSettings` and configure these extra parameters already there. This would in future allow us to have more such parameters out-of-the-box.

zigaLuksic: I suggest we delay generalizing this to all configs until there is another usecase for it. But yes, this was something we already stumbled upon when researching how to pass secrets to eogrow configs.

tests/test_core/test_storage.py
AleksMat: Not really. The way this works is:

1. some profile name is passed from config or env variable to `StorageManager`,
2. `StorageManager` uses an utility that uses `botocore` to try load credentials from `~/.aws/credentials` that are saved under given profile name.
3. Because the profile name doesn't exist we get `ProfileNotFound` exception.
4. The test checks the exception's error message to see if the name of the profile it had to load is the one that we expected it to try to load.

tests/test_core/test_storage.py
AleksMat: Added a couple of sentences of extra explanation into the docstring of the function. Should be clearer now.

PR #89, opened 2022-08-22
Add raster_shape param to rasterize pipeline

zigaLuksic:
Exposes another parameter of the rasterization task.

I also noticed a common pattern of validators, which i managed to extract into `ensure_exactly_one_defined` function. It was tested locally.

### DIFF
--- a/eogrow/pipelines/download.py
+++ b/eogrow/pipelines/download.py
@@ -9,7 +9,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import ray
-from pydantic import Field, root_validator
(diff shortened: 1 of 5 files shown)


zigaLuksic:
> I think it wouldn't hurt if you also added a test for the new validator. I am of the opinion that the util functions that have the potential to be used all over the place should probably be the most foolproof and tested. Other than that, it looks good.

You are right and i was thinking of it. I'll make something happen.

zigaLuksic:
On second thought, i'll add tests in another MR, which will be dedicated to adding tests for all validator utilities we use.

veseln:
In that case, this can be merged.

PR #95, opened 2022-08-31
fix issues found by new version of pydantic

zigaLuksic:
Pydantic updated their mypy plugin in the new version.
This makes it check defaults and default factories. It actually found a few bugs! But mostly I just had to specialize default factories.

### DIFF
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -12,6 +12,7 @@ from eolearn.core.utils.fs import get_aws_credentials, get_filesystem, is_s3_pat
 from sentinelhub import SHConfig
 
 from ..utils.types import AwsAclType
+from ..utils.validators import dict_factory
 from .base import EOGrowObject
 from .schemas import ManagerSchema
 
@@ -40,7 +41,7 @@ class StorageManager(EOGrowObject):
             )
         )
         structure: Dict[str, str] = Field(
-            default_factory=Dict[str, str],
+            default_factory=dict_factory,
             description="A flat key: value store mapping each key to a path in the project.",
         )
 
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -40,7 +40,8 @@ class StorageManager(EOGrowObject):
             )
         )
         structure: Dict[str, str] = Field(
-            default_factory=dict, description="A flat key: value store mapping each key to a path in the project."
+            default_factory=Dict[str, str],
+            description="A flat key: value store mapping each key to a path in the project.",
         )
 
         class Config(ManagerSchema.Config):
--- a/eogrow/pipelines/download.py
+++ b/eogrow/pipelines/download.py
@@ -68,7 +68,7 @@ class RescaleSchema(BaseSchema):
 
 class PostprocessingRescale(BaseSchema):
     rescale_schemas: List[RescaleSchema] = Field(
-        default_factory=List[RescaleSchema], description="Specify different ways to rescale features"
+        default_factory=list_factory, description="Specify different ways to rescale features"
     )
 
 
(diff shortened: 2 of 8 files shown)


eogrow/core/storage.py
```diff
@@ -40,7 +40,8 @@ class Schema(ManagerSchema, BaseSettings):
             )
         )
         structure: Dict[str, str] = Field(
-            default_factory=dict, description="A flat key: value store mapping each key to a path in the project."
+            default_factory=Dict[str, str],
```
mlubej, line 43: Interesting, I though these defaults had to be objects, not types. Does this actually initialize as an empty dictionary, if nothing is provided in some cases like below, where the default is an empty list?

eogrow/pipelines/training.py
```diff
@@ -59,7 +59,7 @@ class Schema(Pipeline.Schema):
         train_test_split: RandomTrainTestSplitSchema
 
         model_parameters: Dict[str, Any] = Field(
-            default_factory=dict, description="Parameters to be provided to the model"
+            default_factory=Dict[str, Any], description="Parameters to be provided to the model"
```
mlubej, line 62: Is such duplication necessary, or does it perhaps automatically infer the default factory from the type, if no default factory is provided?

eogrow/core/storage.py
zigaLuksic: it actually fails, I was hoping it would work. Apparently i'll need to do something else.

eogrow/pipelines/training.py
zigaLuksic: without a default factory the field is required.

eogrow/pipelines/download.py
```diff
@@ -67,7 +68,7 @@ class RescaleSchema(BaseSchema):
 
 class PostprocessingRescale(BaseSchema):
     rescale_schemas: List[RescaleSchema] = Field(
-        default_factory=list, description="Specify different ways to rescale features"
+        default_factory=List[RescaleSchema], description="Specify different ways to rescale features"
```
mlubej, line 71: missed update to `list_factory`?

eogrow/pipelines/training.py
```diff
@@ -183,7 +184,7 @@ class ClassificationPreprocessSchema(BaseSchema):
     )
 
     filter_classes: List[int] = Field(
-        default_factory=list,
+        default_factory=List[int],
```
mlubej, line 187: should be `list_factory`

eogrow/utils/validators.py
```diff
@@ -18,6 +18,16 @@
...
+def list_factory() -> List[Any]:
+    """Required due to changes in pydantic mypy-plugin. Hopefully can be replaced by `list` in future."""
+    return []
+
+
+def dict_factory() -> Dict[Any, Any]:
+    """Required due to changes in pydantic mypy-plugin. Hopefully can be replaced by `dict` in future."""
+    return {}
+
+
```
mlubej, line 30: If I understand correctly, you did it like this so that in future in can be replaced in a single place, right?

eogrow/pipelines/download.py
zigaLuksic: yes, apparently tests dont touch that :see_no_evil:

eogrow/utils/validators.py
zigaLuksic: if this is ever corrected in mypy or pydantic-mypy plugin, then we can delete this and search-replace all occurrences. Also `list_factory` returns `List[Any]` which can then be specialized to whatever type, however `list` returns `List[_T]` and neither mypy nor the pydantic plugin carry out the typevar instantiation, so it fails. Bizzare days we live in

mlubej: 0_0

mlubej (approved):

PR #100, opened 2022-09-28
Add pre commit hooks

zigaLuksic:
Similar to shpy and eo-learn. I had to disable JSON checking because it doesn't play nicely with comments. I also disabled SIM117 (merging of  `with` statements) because with statements get large in eo-grow, but we dont have 3.10 for the syntax improvements.


eogrow/core/area/base.py
```diff
@@ -301,4 +301,4 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
         """Adds a column with bounding boxes to all dataframes in a grid"""
         for bbox_df in grid:
             crs = CRS(bbox_df.crs)
-            bbox_df["BBOX"] = bbox_df.geometry.apply(lambda geo: BBox(geo.bounds, crs))
+            bbox_df["BBOX"] = bbox_df.geometry.apply(lambda geo: BBox(geo.bounds, crs))  # noqa B023
```
mlubej, line 304: > B023: Functions defined inside a loop must not use variables redefined in the loop, because [late-binding closures are a classic gotcha](https://docs.python-guide.org/writing/gotchas/#late-binding-closures).

is this just because there is a lambda function here?

eogrow/pipelines/sampling.py
```diff
@@ -154,7 +154,7 @@ def _get_output_features(self) -> List[FeatureSpec]:
         return output_features
 
 
-class BaseRandomSamplingPipeline(BaseSamplingPipeline, metaclass=abc.ABCMeta):
+class BaseRandomSamplingPipeline(BaseSamplingPipeline, metaclass=abc.ABCMeta):  # noqa B024
```
mlubej, line 157: This one is above my pay-grade

eogrow/tasks/features.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, validity_threshold: float):
 
     def __call__(self, array: np.ndarray) -> bool:
         coverage = np.sum(array.astype(np.uint8)) / np.prod(array.shape)
-        return bool(coverage > self.validity_threshold)  # numpy return np.bool_ by default
+        return bool(coverage > self.validity_threshold)  # noqa SIM901 # numpy return np.bool_ by default
```
mlubej, line 31: what about 

```python
(coverage > self.validity_threshold).astype(bool)
```

or dropping the `numpy` layer with

```python
coverage = (np.sum(array.astype(np.uint8)) / np.prod(array.shape)).astype(float)
```

eogrow/utils/batch.py
```diff
@@ -6,7 +6,7 @@
 
 def read_timestamps_from_orbits(userdata: dict) -> List[str]:
     """Parses batch orbits payload to obtain a list of timestamp strings"""
-    if "orbits" in userdata:
+    if "orbits" in userdata:  # noqa SIM908
```
mlubej, line 9: > [SIM908](https://github.com/MartinThoma/flake8-simplify/issues/50): Use dict.get(key) ([example](https://pypi.org/project/flake8-simplify/#SIM908))

what a stupid flake8 simplification, why not ignore it all-together?

.flake8
```diff
@@ -1,8 +1,9 @@
 [flake8]
-ignore = E203, W503
+ignore = E203, W503, C408, SIM117
```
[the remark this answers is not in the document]
zigaLuksic: C408 is ignored because otherwise it insists that `dict(a=2, b=3)` is rewritten as `{"a": 2, "b":3}`. In long dictionaries the first option is less verbose because you avoid all the double quotes, so we ignore it in all eo-* packages.

SIM117:
```python
with LocalFile(...) as local:
    with open(...) as f:
        ...

# wants it to be changed to
with LocalFile(...) as local, open(...) as f:
    ...
```
This indeed gets rid of identation levels, BUT in eo-grow we have huge rasterio context managers, which span multiple lines. These cannot be merged without disgusting syntax (until 3.10).

eogrow/core/area/base.py
zigaLuksic: yes. It is not dangerous in our case because it is applied instantly, but python closures are horrifying.

eogrow/pipelines/sampling.py
zigaLuksic: this has to do with ABCs and inheritance. If you're inheriting from an abstract class, but wish to remain abstract (i.e. not implement the methods, just add some functionalities), flake8 doesn't realize that your class still has some abstract methods (inherited as abstract and never implemented) and marks it as a false abstract class

eogrow/tasks/features.py
zigaLuksic: the first one works :D it's longer than the `bool` version but ultimately shorter as it gets rid of the comment

eogrow/utils/batch.py
zigaLuksic: because in most cases it is better to use `get`. The issue is how we're doing the whole *you are you, unless you're a dictionary in which case you are your subitem*, which is just bad practice.

mlubej: Why not then fix the problem at the source? I'd be happy to lose a bad practice habit. Or is this case an exception?

eogrow/tasks/features.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, validity_threshold: float):
 
     def __call__(self, array: np.ndarray) -> bool:
         coverage = np.sum(array.astype(np.uint8)) / np.prod(array.shape)
-        return bool(coverage > self.validity_threshold)  # numpy return np.bool_ by default
+        return (coverage > self.validity_threshold).astype(bool)
```
mlubej, line 31: I remember some fixes in `numpy` where `numpy` changed `np.bool` to `bool`. Could it be that the `numpy` version is some older one and this wouldn't be necessary with a newer version in the first place?

zigaLuksic: god knows, we are stuck with the old version due to `numba`

eogrow/utils/batch.py
zigaLuksic: I'd say this is an exception (for now) due to how the timestamps reader is done

PR #101, opened 2022-09-30
Fix setting of nodata in export2tiff pipeline, round 2

mlubej:
ExportToTiff pipeline didn't behave as expected. When merging tiffs, I wanted to set the empty space to the no-data value, turns out there were issues because of:
- not setting the `-init` param
- not using the values as strings (- was understood as a parameter???)

Thanks @batic for the help.

### DIFF
--- a/eogrow/utils/map.py
+++ b/eogrow/utils/map.py
@@ -142,7 +142,7 @@ def merge_tiffs(
     gdalmerge_options = "-co BIGTIFF=YES -co compress=LZW"
 
     if nodata is not None:
-        gdalmerge_options += f" -a_nodata {nodata}"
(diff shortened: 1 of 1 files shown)


mlubej:
> is there a way to add a test for this?

One should be able to read metainfo from the raster and compare with the expected values for various combinations of parameters and parameter values.

zigaLuksic:
> > is there a way to add a test for this?
> 
> One should be able to read metainfo from the raster and compare with the expected values for various combinations of parameters and parameter values.

are you up for it? otherwise just open an issue on the internal repo

zigaLuksic (approved):

PR #104, opened 2022-10-06
Refactor ExportMap pipeline to allow easier changes in future

zigaLuksic:
I tried adding support for temporal byocs, but the pipeline itself was a bit hard to read. So as a first step I'd like to reorganize it a bit. Should make later changes easier to follow.

Seems like github butchered the diffs a bit... Its mostly reordering and extracting to methods.

### DIFF
--- a/eogrow/pipelines/export_maps.py
+++ b/eogrow/pipelines/export_maps.py
@@ -69,28 +69,6 @@ class ExportMapsPipeline(Pipeline):
 
         return successful, failed
 
-    def _merge_maps(self, input_filename_list: List[str], merged_filename: str) -> None:
-        """Performs gdal_merge on a set of given geotiff images, make them pixel aligned cogs if `cogify` is True
-
-        :param input_filename_list: A list of input tiff image filenames
-        :param merged_filename: Filename of merged tiff image
-        :param blocksize: block size of tiled tiff
-        """
-
-        merge_tiffs(
-            input_filename_list,
-            merged_filename,
-            overwrite=True,
-            delete_input=False,
-            nodata=self.config.no_data_value,
-            dtype=self.config.map_dtype,
-        )
-
-        if self.config.cogify:
-            cogify_inplace(
-                merged_filename, blocksize=1024, nodata=self.config.no_data_value, dtype=self.config.map_dtype
-            )
-
     def build_workflow(self) -> EOWorkflow:
         """Method where workflow is constructed"""
         load_task = LoadTask(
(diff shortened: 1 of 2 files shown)


eogrow/pipelines/export_maps.py
```diff
@@ -18,7 +18,7 @@
 from eolearn.io import ExportToTiffTask
 
 from ..core.pipeline import Pipeline
-from ..utils.map import merge_maps
+from ..utils.map import cogify_inplace, merge_tiffs
```
mlubej, line 21: I guess this means the `merge_maps` can be deleted in the utils? Perhaps you forgot to push.

eogrow/pipelines/export_maps.py
```diff
@@ -114,16 +128,46 @@ def create_maps(self, feature_name: str, patch_names: List[str]) -> None:
...
+        merged_map_path: str,
+    ) -> Tuple[Optional[TempFS], List[str], str]:
+        """Returns system paths that can be used to merge maps.
+
+        If required files are copied locally and a temporary filesystem object is returned.
+        """
+        make_local_copies = self.storage.is_on_aws() or self.config.force_local_copies
+
         if make_local_copies:
             temp_fs = TempFS(identifier="_merge_maps_temp")
```
mlubej, line 172: If I remember correctly, the issue in the past was that this merging took a reeeeeally long time if a lot of tiffs were being merged from S3, so in that case it was better to first copy the tiffs to a local dir, then merge, and then copy to the target place. 

If I understand correctly, this is now not the case and could potentially be problematic again, right? (assuming I correctly remember the issue)

eogrow/pipelines/export_maps.py
zigaLuksic: I was hesitant with doing this, because 'someone might be using it somewhere' but i think you're right, it should be deleted, since it's just two lines anyway.

eogrow/pipelines/export_maps.py
zigaLuksic: That is still a potential problem. If i did everything correctly then the files should still be copied when needed.

mlubej: Ah, I missed the part where you put this into a specific function. The logic seems to stay the same, yes.

mlubej (approved):

eogrow/pipelines/export_maps.py
mlubej: Feel free to resolve and merge after update.

PR #107, opened 2022-10-11
Improve tests for eogrow.utils

zigaLuksic:
@mlubej You said that we have some bug for the cogification of uint8 right? Can you tell me how to add a test that would detect that?

### DIFF
--- a/tests/test_utils/test_map.py
+++ b/tests/test_utils/test_map.py
@@ -1,19 +1,21 @@
 """
 Tests for utils.map module
 """
-from typing import Optional
+from typing import List, Optional
 
 import numpy as np
 import pytest
 import rasterio
 from fs.base import FS
 from fs.tempfs import TempFS
+from numpy.testing import assert_array_almost_equal
+from pytest import approx
 
 from eolearn.core import EOPatch, FeatureType
 from eolearn.io import ExportToTiffTask
 from sentinelhub import CRS, BBox
 
-from eogrow.utils.map import GDAL_DTYPE_SETTINGS, cogify, cogify_inplace
+from eogrow.utils.map import GDAL_DTYPE_SETTINGS, cogify, cogify_inplace, merge_tiffs
 
 pytestmark = pytest.mark.fast
 
@@ -39,7 +41,7 @@ class TestCogify:
         make_geotiff(data, BBox((1, 2, 3, 4), CRS.WGS84), filesystem, name="input")
         return filesystem.getsyspath("input.tif")
 
-    @pytest.fixture(name="output_path")
+    @pytest.fixture(name="output_path", scope="class")
     def output_file_fixture(_, filesystem: FS) -> str:
         return filesystem.getsyspath("output.tif")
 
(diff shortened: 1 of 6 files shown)


mlubej:
> You said that we have some bug for the cogification of uint8 right? Can you tell me how to add a test that would detect that?

I would rephrase my initial words into:
- never use `(u)int8` for exporting maps, use what [gdal supports officially](https://gdal.org/drivers/raster/gtiff.html)
  - `Byte, UInt16, Int16, UInt32, Int32, Float32, Float64, CInt16, CInt32, CFloat32 and CFloat64`
  - probably this only makes sense for `UInt16`, `Int16`, `UInt32`, `Int32`, `Float32`, `Float64`
- the `no_data` provided when exporting should also be the value defined for regions outside of the processed AOI

The issue was that `(u)int8` was actually a byte or a signed byte, but it didn't work as expected, so it's best to throw it out. Perhaps `bool` could be used as a mapping for `Byte` type

zigaLuksic:
> > You said that we have some bug for the cogification of uint8 right? Can you tell me how to add a test that would detect that?
> 
> I would rephrase my initial words into:
> 
>     * never use `(u)int8` for exporting maps, use what [gdal supports officially](https://gdal.org/drivers/raster/gtiff.html)
>       
>       * `Byte, UInt16, Int16, UInt32, Int32, Float32, Float64, CInt16, CInt32, CFloat32 and CFloat64`
>       * probably this only makes sense for `UInt16`, `Int16`, `UInt32`, `Int32`, `Float32`, `Float64`
> 
>     * the `no_data` provided when exporting should also be the value defined for regions outside of the processed AOI
> 
> 
> The issue was that `(u)int8` was actually a byte or a signed byte, but it didn't work as expected, so it's best to throw it out. Perhaps `bool` could be used as a mapping for `Byte` type

I'll throw it out with the force of 50 gorillas, but first I need a test that will show me it's wrong (so that we don't add it back in in 2 months time when my short-term-memory refreshes)

mlubej:
> > > You said that we have some bug for the cogification of uint8 right? Can you tell me how to add a test that would detect that?
> > 
> > 
> > I would rephrase my initial words into:
> > ```
> > * never use `(u)int8` for exporting maps, use what [gdal supports officially](https://gdal.org/drivers/raster/gtiff.html)
> >   
> >   * `Byte, UInt16, Int16, UInt32, Int32, Float32, Float64, CInt16, CInt32, CFloat32 and CFloat64`
> >   * probably this only makes sense for `UInt16`, `Int16`, `UInt32`, `Int32`, `Float32`, `Float64`
> > 
> > * the `no_data` provided when exporting should also be the value defined for regions outside of the processed AOI
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The issue was that `(u)int8` was actually a byte or a signed byte, but it didn't work as expected, so it's best to throw it out. Perhaps `bool` could be used as a mapping for `Byte` type
> 
> I'll throw it out with the force of 50 gorillas, but first I need a test that will show me it's wrong (so that we don't add it back in in 2 months time when my short-term-memory refreshes)

More concretely, if the following was true:
- tif contained 0 and 1 
- tiff had -1 as `no_data`
- tif was saved as an `int8` ([signed byte in the background](https://github.com/sentinel-hub/eo-grow/blob/5c0410bb0df6ee108b0f0471bf473fb674cae162/eogrow/utils/map.py#L16))

then the `no_data` was not actually -1. I don't have any concrete cases at hand, though. But I think this is really an obscure case where we tried to hack our way into using `(u)int8)`. I believe if you just throw it out and throw an error whenever someone wants to save a `(u)int8` tiff, then this should never happen in a billion years.

tests/test_utils/test_map.py
```diff
@@ -0,0 +1,134 @@
...
+        return np.stack(layers, axis=-1).astype(np.int32)
+
+    @pytest.fixture(name="input_path", scope="class")
+    def input_file_fixture(_, input_data: np.ndarray, filesystem: FS) -> str:
+        make_geotiff(input_data, BBox((1, 2, 3, 4), CRS.WGS84), filesystem, name="input")
+        return filesystem.getsyspath("input.tif")
+
+    @pytest.mark.parametrize("bands", [[0], [1], [2], [0, 1], [2, 1, 0]])
+    def test_extract_bands(self, output_path: str, input_path: List[str], input_data: np.ndarray, bands: List[int]):
+        extract_bands(input_path, output_path, bands, overwrite=True)
```
jgersak, line 130: What happens with an empty list do we have an exception and should we test it ?

zigaLuksic: if the list is empty then all of the bands are transferred over. Unlikely to be passed as an input, but we probably should raise an error in that case.

jgersak (approved):

PR #108, opened 2022-10-14
Byoc friendly export of temporal features

zigaLuksic:
Adds an option to temporally split maps when exporting. This will make it much simpler to do byoc ingestion.

I also resolved some path-juggling by enforcing the rule that all paths are relative to the filesystem used (storage or tempfs) and system paths are used only when calling the gdal functions 

Each map is suffixed with the timestamp in a near-iso format (the `:` character is problematic on windows so I tried to avoid it).

The output result of the test is rather huge (3k+ lines) due to around 30 timestamps... I would want it to be run as part of the test-chain that is done on github, and for that I cannot give it less data. But perhaps we could just run a smaller pipeline in the chain-test and have this only as part of the large test suite? Then i can switch it over to some smaller data as well.


mlubej:
> another question, do you think it would make more sense to switch from `folder/UTM_something/blabla_TIMESTAMP.tif` to `folder/UTM_something/TIMESTAMP/blabla.tif` ? I guess it would be better for ingesting multi-band BYOC collections where each band is a separate tiff, but i'm not sure if multiband tiffs and single-bands tiffs play well together


It might be easier for ingesting, yes, especially if there are multiple tiffs for the same timestamps, then the dir structure is a bit more intuitive.

> I'm mostly worried because tiff names reflect in the band names of the collection, so having timestamps in band names would be odd.

From ingestion point of view this makes sense. From export-only point of view I don't think it changes much. Temporal export is rarely used anyway, so it shouldn't be an invasive change.

eogrow/pipelines/export_maps.py
```diff
@@ -56,6 +61,13 @@ class Schema(Pipeline.Schema):
                 " With this parameter you force to always make copies."
             ),
         )
+        compress_temporally: bool = Field(
```
mlubej, line 64: This might be a bit confusing for the user. IIRC the output is a mega-tiff (a multi-band tiff containing bands and timestamps in a single tiff). From that perspective, the parameter name `compress_temporally` might be a bit confusing, especially since in the description below it says `Performs temporal split of tiffs if needed`.

I'm a bit confused as to how exactly the output should look, so I cannot give a proper suggestion, just a warning that I find the name confusing.

eogrow/pipelines/export_maps.py
```diff
@@ -76,54 +90,61 @@ def run_procedure(self) -> Tuple[List[str], List[str]]:
...
+            output_paths: List[Tuple[str, Optional[dt.datetime]]]
+            if feature_type.is_timeless() or self.config.compress_temporally:
+                output_paths = [(map_path, None)]
+            else:
+                timestamp = self._load_timestamp(eopatch_list[0])  # we assume all eopatches share the same timestamp
+                output_paths = self._split_temporally(filesystem, map_path, timestamp, output_folder)
 
-            self._finalize_output_files(temp_fs, map_sys_path, merged_map_path)
+            if self.config.cogify:
+                resampling = "mode" if feature_type.is_discrete() else "bilinear"
```
mlubej, line 124: is this stable? can it happen that you have `[0.0, 1.0, 4.0, 10.0]` values which are actually discrete but still have a floating point? 

I don't know if it makes sense to solve for such cases anyway

eogrow/pipelines/export_maps.py
zigaLuksic: I updated the description a bit.
By default you get multiple tiffs (one per timestamp) but with this enabled you get a big chungus, which has `bands * timestamps` bands (the old behaviour)

eogrow/pipelines/export_maps.py
zigaLuksic: This uses `mode` for `FeatureType.MASK` and `FeatureType.MASK_TIMELESS`. Those two feature types are also checked to be discrete. The only other way of doing this would be to check the dtype at runtime i guess. Feels a bit hacky

mlubej: 👍

eogrow/pipelines/export_maps.py
```diff
@@ -56,6 +61,13 @@ class Schema(Pipeline.Schema):
                 " With this parameter you force to always make copies."
             ),
         )
+        compress_temporally: bool = Field(
+            False,
+            description=(
+                "Temporal features are by default exported as multiple TIFF files in a per-timestamp manner. This "
+                "parameter results in a single TIFF file with the same band order as the one of `ExportToTiffTask`."
```
mlubej, line 68: Enabling this parameter results in *

mlubej (approved):
One more comment for the description fix, otherwise good to go.

PR #109, opened 2022-10-14
Byoc pipeline tests

zigaLuksic:
It's scary to add functionality to a pipeline that is not tested properly, so I added some base tests before upgrading

### DIFF
--- a/eogrow/pipelines/byoc.py
+++ b/eogrow/pipelines/byoc.py
@@ -60,7 +60,7 @@ class IngestByocTilesPipeline(Pipeline):
         if not self.storage.is_on_aws:
             raise ValueError("Can only ingest for projects based on S3 storage.")
         project_folder = self.storage.config.project_folder
-        self.bucket_name = project_folder.split("/")[2]
+        self.bucket_name = project_folder.replace("s3://", "").split("/")[0]
         self._cover_geometry_df: Optional[gpd.GeoDataFrame] = None
 
     def get_byoc_collection(self, byoc_client: SentinelHubBYOC) -> JsonDict:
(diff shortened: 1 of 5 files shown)


eogrow/pipelines/byoc.py
```diff
@@ -101,7 +101,7 @@ def _get_tiff_paths(self) -> List[str]:
     def _get_byoc_compliant_path(self, relative_path: str) -> str:
         """Transforms a project relative path to a bucket-name relative path that is required by BYOC."""
         absolute_path = fs.path.combine(self.config.storage.project_folder, relative_path)  # type: ignore[attr-defined]
-        return absolute_path[6 + len(self.bucket_name) :]  # removes s3://<bucket-name>/
+        return absolute_path.replace(f"s3://{self.bucket_name}", "")  # removes s3://<bucket-name>/
```
mlubej, line 104: Perhaps unclear comment, since this doesn't remove the trailing fwd slash?

tests/test_pipelines/test_byoc.py
```diff
@@ -0,0 +1,99 @@
...
+def test_timeless_byoc(config_folder, preparation_config, config, requests_mock):
+    """Requires heavy mocking to get any testing done.
+
+    Mocks:
+    - The pipeline has to be tricked into thinking it's on AWS during init so it doesn't raise an error.
+    - The bucket_name is changed since it's not parsed correctly (because we're not on S3).
+    - When reading the cover geometry the path is not filesystem relative (but a join of bucket name + bucket relative)
+      so we mock it to prevent `rasterio.open` to fail.
+    - All request endpoints are mocked in the `requests_mock` fixture.
+    """
```
mlubej, line 62: Perhaps it makes sense to create a low-priority ticket to investigate usage of `moto` for some future time? I feel it might be more elegant, but I'm aware there were some issues in the past. Perhaps they've been resolved? 

In any case, it's not the point of this MR.

eogrow/pipelines/byoc.py
zigaLuksic: great catch <3 The comment is correct, my change is not. It should remove the slash as well but I can't count for shit.

tests/test_pipelines/test_byoc.py
zigaLuksic: yes, the whole suite should be 'double-tested', once locally and once in a mocked S3 environment. I'll add another issue for the tests

PR #110, opened 2022-10-18
Ingest temporal byoc

zigaLuksic:
1. makes the pipeline work with temporal byocs
2. fixes a filesystem issue for exporting tiffs (found when testing) and adapts test to detect that
3. improves test suite for byoc pipeline and adds the byoc tests to chain (as they are only 3-5s)


eogrow/pipelines/byoc.py
```diff
@@ -27,25 +28,40 @@
 )
 
 LOGGER = logging.getLogger(__name__)
+TIMESTAMP_FORMAT = "%Y-%m-%dT%H-%M-%S"  # TODO: import from .export_maps once MRs are synced
```
mlubej, line 31: Why not just change the source/rebase branch to the other branch, this way the changes would automatically be included.

eogrow/pipelines/byoc.py
```diff
@@ -27,25 +28,40 @@
...
-        file_glob_pattern: str = Field("**/UTM_*/*.tiff", description="Pattern used to obtain the TIFF files to use")
+        file_glob_pattern: str = Field("**/*.tiff", description="Pattern used to obtain the TIFF files to use")
 
         new_collection_name: Optional[str] = Field(description="Used for defining a new BYOC collection.")
         existing_collection: Optional[DataCollection] = Field(description="Used when updating and reingesting.")
         _parse_byoc_collection = optional_field_validator("existing_collection", parse_data_collection, pre=True)
         _ensure_exclusion = ensure_exactly_one_defined("new_collection_name", "existing_collection")
 
-        sensing_time: datetime.datetime = Field(description="Sensing time (ISO format) added to BYOC tiles.")
+        is_temporal: bool = Field(
```
mlubej, line 46: How is the timestamp from the parent folder inferred? should the user follow some specific instruction on how to produce the storage structure to successfully infer the timestamps?

eogrow/pipelines/byoc.py
```diff
@@ -27,25 +28,40 @@
...
 
         @validator("sensing_time", pre=True)
-        def _parse_sensing_time(cls, value: str) -> datetime.datetime:
-            return datetime.datetime.fromisoformat(value)
+        def _parse_sensing_time(cls, value: Optional[str], values: Dict[str, object]) -> Optional[datetime.datetime]:
+            is_temporal = values["is_temporal"]
+            if is_temporal and value is None:
+                return None
+            if not is_temporal and value is not None:
+                return datetime.datetime.fromisoformat(value)
```
mlubej, line 63: does this allow providing just `YYYY-MM-DD`? If not, that would be quite convenient.

eogrow/pipelines/byoc.py
```diff
@@ -109,7 +125,14 @@ def _prepare_tile(self, folder: str, tiff_paths: List[str]) -> Optional[ByocTile
         cover_geometry = self._get_tile_cover_geometry(some_tiff)
         if cover_geometry.geometry.is_empty:
             return None
-        return ByocTile(folder, cover_geometry=cover_geometry, sensing_time=self.config.sensing_time)
+
+        sensing_time = self.config.sensing_time
+        if self.config.is_temporal:
+            # assumes file-system structure is equal to that of ExportMapsPipeline
+            timestamp_folder = fs.path.split(folder)[1]
+            sensing_time = dt.datetime.strptime(timestamp_folder, TIMESTAMP_FORMAT)
```
mlubej, line 133: Since the user is the one who provides the BYOCs, shouldn't the user be able to define the timestamp format? or do you want to limit the complexity of what can be done? In that case I would make it more explicit as to what the structure is.

eogrow/pipelines/byoc.py
zigaLuksic: because this was done before that branch was merged and it would break the tests
cleaned now

eogrow/pipelines/byoc.py
zigaLuksic: it does, see [here](https://github.com/sentinel-hub/eo-grow/blob/develop/tests/test_config_files/byoc/ingest_lulc.json)

eogrow/pipelines/byoc.py
zigaLuksic: just gonna answer it here since this is similar to the comment above. The pipeline currently works mostly in conjunction with the `ExportMapsPipeline`. I guess i can document the structure more, but i'm not sure where. At this point i don't feel like this should be too configurable, once we have a use-case for it we can adjust if needed.

mlubej: Perhaps it would make sense to add a short sentence in the description of the `is_temporal` parameter with an example of what is expected, just to make it more intuitive. 

In any case, not so important, feel free to ignore and resolve.

zigaLuksic: added an example

PR #114, opened 2022-10-20
Expose resampling for overviews in cogification process

zigaLuksic:


eogrow/utils/map.py
```diff
@@ -64,6 +69,8 @@ def cogify(
     :param nodata: value to be treated as nodata, default value is None
     :param dtype: output type of the in the resulting tiff, default is None
     :param overwrite: If True overwrite the output file if it exists.
+    :param resampling: The resampling method used to produce overviews. The defaults (when using None) are CUBIC for
```
mlubej, line 72: I'm not sure this is used anywhere.

zigaLuksic: what you actually meant to say is "Hey you dork, you forgot to delete line 85, get your shit together"

fixed

zigaLuksic: i'll also deal with the defaults in a second

zigaLuksic: the resampling is now not passed if it's set to `None` to let the defaults work their magic.

mlubej (approved):

PR #115, opened 2022-10-21
Add warp resampling when merging tiffs.

zigaLuksic:
By switching from `gdal_merge` to `gdalwarp` we can now specify how to resample tiffs that are warped. This is a possible improvement for pixel misalignment.

Another benefit is that `gdal_merge` loads all files into memory while `gdalwarp` runs much more memory-conservative.

### DIFF
--- a/eogrow/pipelines/export_maps.py
+++ b/eogrow/pipelines/export_maps.py
@@ -22,7 +22,7 @@ from eolearn.features import LinearFunctionTask
 from eolearn.io import ExportToTiffTask
 
 from ..core.pipeline import Pipeline
-from ..utils.map import cogify_inplace, extract_bands, merge_tiffs
+from ..utils.map import CogifyResamplingOptions, WarpResamplingOptions, cogify_inplace, extract_bands, merge_tiffs
 from ..utils.types import Feature
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


mlubej:
I wanted to understand more the differences between `gdal_merge.py` and `gdal_warp` and found the following statements, linking here if anyone else wants a TL;DR:

> gdal_merge.py loads all files into memory before processing them. t[here](http://osgeo-org.1560.x6.nabble.com/gdal-merge-MemoryError-td3775927.html)fore it is not able to process large files if your memory is small.

> gdal_merge.py uses nearest neighbor resampling. If you want control over the resampling used, you should use gdalwarp instead.

mlubej:
Also found the following hint for a potential speed up:

> While I don't know why GDAL provides this overlap in functionality, be sure to set the cache for gdalwarp to make it really fast:

```bash
 # assuming 3G of cache here: 
 gdalwarp --config GDAL_CACHEMAX 3000 -wm 3000 $(list_of_tiffs) merged.tiff
```

zigaLuksic:
> Also found the following hint for a potential speed up:
> 
> > While I don't know why GDAL provides this overlap in functionality, be sure to set the cache for gdalwarp to make it really fast:
> 
> ```shell
>  # assuming 3G of cache here: 
>  gdalwarp --config GDAL_CACHEMAX 3000 -wm 3000 $(list_of_tiffs) merged.tiff
> ```
> 
> from: https://gis.stackexchange.com/a/44739

I found this as well, but i'm not entirely sure how to do that. Apparently in cases of lots of small tiffs this can actually be detrimental. I guess exposing another parameter? But perhaps i'd wait to see if this is ever a bottleneck

mlubej (approved):

PR #116, opened 2022-10-24
Remove gdal utility tests from github-suite as they fail on schedule (but not on PR or push)

zigaLuksic:


zigaLuksic:
after further investigation, the issue should resolve itself after a release

PR #119, opened 2022-10-26
Improvements for ExportToTiff pipeline

zigaLuksic:
1. Removes disk-space issue that occurs due to `extract_bands` utility function not cleaning up after copying. The copying was deemed not necessary, since the utility function doesn't get used in-place (and if it does the user should take care of copying)
2. Local eo-patch tiffs are removed ASAP to preserve disk space
3. Tiffs are exported with compression. Makes the pipeline a bit slower, but saves on space (50% reduction in file-size). The names are also no longer unique-per-run, so subsequent runs override the previous ones (if the previous pipeline failed), reducing the clutter on storage when debugging.

### DIFF
--- a/eogrow/pipelines/export_maps.py
+++ b/eogrow/pipelines/export_maps.py
@@ -182,6 +182,7 @@ class ExportMapsPipeline(Pipeline):
             no_data_value=self.config.no_data_value,
             image_dtype=np.dtype(self.config.map_dtype),
             band_indices=self.config.band_indices,
+            compress="LZW",
         )
         task_list.append(export_to_tiff_task)
 
@@ -199,7 +200,10 @@ class ExportMapsPipeline(Pipeline):
 
     def get_geotiff_name(self, name: str) -> str:
         """Creates a unique name of a geotiff image"""
-        return f"{name}_{self.__class__.__name__}_{self.pipeline_id}.tiff"
+        return f"{name}_{self._get_map_name()}.tiff"
+
+    def _get_map_name(self) -> str:
+        return self.config.map_name or f"{self.config.feature[1]}.tiff"
 
     def _prepare_files(self, geotiff_paths: List[str], output_file: str) -> Tuple[FS, List[str], str]:
         """Returns system paths of geotiffs and output file that can be used to merge maps.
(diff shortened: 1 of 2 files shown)


eogrow/pipelines/export_maps.py
```diff
@@ -195,9 +199,18 @@ def get_execution_arguments(self, workflow: EOWorkflow) -> List[Dict[EONode, Dic
...
+        base = f"{self._get_map_name(with_extension=False)}_{name}"
+        if crs is not None:
+            base += f"_UTM_{crs.epsg}"
+        if time is not None:
+            base += f"_{time.strftime(TIMESTAMP_FORMAT)}"
+        return f"{base}.tiff"
+
+    def _get_map_name(self, with_extension: bool = True) -> str:
+        name = self.config.map_name or f"{self.config.feature[1]}.tiff"
+        return name if with_extension else name.replace(".tiff", "")
```
mlubej, line 213: perhaps using `os.path.split()[0]` or some variation of it would be better? Not sure how much variation is there in the tiff name, since sometimes it's `.tif` and sometimes `.tiff`, but not sure if it's hard-coded here or not

eogrow/pipelines/export_maps.py
```diff
@@ -102,15 +103,14 @@ def run_procedure(self) -> Tuple[List[str], List[str]]:
...
-        # TODO: This could be parallelized per-crs
         for crs, eopatch_list in crs_eopatch_dict.items():
             LOGGER.info("Processing for UTM %d", crs.epsg)
 
             output_folder = fs.path.join(folder, f"UTM_{crs.epsg}")
             # manually make subfolder, otherwise things fail on S3 in later steps
             self.storage.filesystem.makedirs(output_folder, recreate=True)
 
-            merged_map_path = fs.path.join(output_folder, self.get_geotiff_name("full_merged_map"))
+            merged_map_path = fs.path.join(output_folder, self.get_geotiff_name("merged", crs=crs))
```
veseln, line 113: I prefer the constant strings to be defined as class constants. But this is just a stylistic preference, you can leave it as is if you wish.

mlubej (approved):

eogrow/pipelines/export_maps.py
veseln: I think there is validation of the input to always be `.tiff` (I checked cause I had a similar concern as you have)

https://github.com/sentinel-hub/eo-grow/blob/ecbd26ba3114c251343afb11300110d8d48c937f/eogrow/pipelines/export_maps.py#L48

zigaLuksic: yeah, fixing it seemed like a good idea at the time
still does :) but I'll switch to the mimetype constant that we have, just to make sure it always matches (which is what Nejc also suggested for map names and he is right)

eogrow/pipelines/export_maps.py
```diff
@@ -123,15 +123,17 @@ def run_procedure(self) -> Tuple[List[str], List[str]]:
                 nodata=self.config.no_data_value,
                 dtype=self.config.map_dtype,
                 warp_resampling=self.config.warp_resampling,
-                quiet=True,
             )
 
+            LOGGER.info("Remove local per-eopatch tiffs after merge.")
+            parallelize(filesystem.remove, geotiff_paths, workers=None, multiprocess=False, desc="Remove tiffs")
```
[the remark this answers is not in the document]
zigaLuksic: it is actually only multithreaded (check the `multiprocess=False`) :D

eogrow/pipelines/export_maps.py
zigaLuksic: done

PR #121, opened 2022-10-28
Update after geopandas 0.12 changes

zigaLuksic:
The eopatch manager does some shady stuff when the area uses multiple CRS.
Geopandas started complaining in 0.12, so this is a quick fix (with a test) that this should no longer be an issue.

### DIFF
--- a/eogrow/core/eopatch.py
+++ b/eogrow/core/eopatch.py
@@ -60,7 +60,7 @@ class EOPatchManager(EOGrowObject):
         """
         bbox_grid = self._area_manager.get_grid(add_bbox_column=True)
 
-        bbox_df: DataFrame = pandas.concat(bbox_grid, ignore_index=True)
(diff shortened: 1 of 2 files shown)


eogrow/core/eopatch.py
```diff
@@ -60,7 +60,7 @@ def _prepare_names(self) -> Tuple[bidict, Dict[str, BBox]]:
         """
         bbox_grid = self._area_manager.get_grid(add_bbox_column=True)
 
-        bbox_df: DataFrame = pandas.concat(bbox_grid, ignore_index=True)
+        bbox_df: DataFrame = pandas.concat([gdf.drop(columns="geometry") for gdf in bbox_grid], ignore_index=True)
```
batic, line 63: It was a pandas `DataFrame` even before, so I assume we've never really needed `geopandas` functionality on the bbox functionality, so I agree with the approach. 

Another - if "geometries" would be needed - would be to convert them to `sentinelhub.Geometry` objects.

batic (approved):

PR #124, opened 2022-11-08
quick cleanup of storage manager

zigaLuksic:
Improved the docstrings just a bit, but the main part was the removal of `show_folder_structure`.
This was useful in LCMS when the storage structure was very complex. I don't remember using this or seeing anyone use this since eogrow 1.0 was released (hence requesting @mlubej for his opinion).

### DIFF
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -1,8 +1,7 @@
 """
 This module handles everything regarding storage of the data
 """
-from io import StringIO
-from typing import Dict, List, Optional
+from typing import Dict, Optional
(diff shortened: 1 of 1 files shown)


mlubej:
> This was useful in LCMS when the storage structure was very complex. I don't remember using this or seeing anyone use this since eogrow 1.0 was released (hence requesting @mlubej for his opinion).

TBH, I never used it in LCMS either, I always resorted to tree view in file browsers etc, so I agree with the action.

mlubej (approved):

PR #125, opened 2022-11-09
Add option of suffix for temporary files in ExportMapsPipeline

zigaLuksic:
After debating on how this is an edge-case and wont happen, we had it happen not even a week later.

For instance, if our time spans multiple years, we could have eopatches saved in a per-year to reduce memory consumption. But then we cant use two clusters, one for exporting 2020 and one for 2021, because the maps have the same name, and the exported tiffs will override eachother (since only the map name matters).

For that edge-case the user can now provide a suffix.

### DIFF
--- a/eogrow/pipelines/export_maps.py
+++ b/eogrow/pipelines/export_maps.py
@@ -74,11 +74,11 @@ class ExportMapsPipeline(Pipeline):
                 " With this parameter you force to always make copies."
             ),
         )
-        temporary_files_suffix: str = Field(
+        interim_results_suffix: str = Field(
             "",
             description=(
-                "Added as a suffix to names of partially complete files in order to avoid clashes from multiple"
-                " clusters working on similar maps (e.g. exporting same map for different timestamps)."
+                "Has no effect on end results. Adds a suffix to names of temporary files in order to avoid clashes"
+                " from pipelines working in parallel on same maps (e.g. exporting same map for different timestamps)."
             ),
         )
         split_per_timestamp: bool = Field(
@@ -103,7 +103,7 @@ class ExportMapsPipeline(Pipeline):
         super().__init__(config, raw_config)
 
         self.map_name = self.config.map_name or f"{self.config.feature[1]}.{MimeType.TIFF.extension}"
-        self.get_tiff_name = partial(get_tiff_name, self.map_name, suffix=self.config.temporary_files_suffix)
+        self.get_tiff_name = partial(get_tiff_name, self.map_name, suffix=self.config.interim_results_suffix)
 
     def run_procedure(self) -> Tuple[List[str], List[str]]:
         """Extracts and merges the data from EOPatches into a TIFF file.
--- a/eogrow/pipelines/export_maps.py
+++ b/eogrow/pipelines/export_maps.py
@@ -74,6 +74,13 @@ class ExportMapsPipeline(Pipeline):
                 " With this parameter you force to always make copies."
             ),
         )
+        temporary_files_suffix: str = Field(
+            "",
+            description=(
+                "Added as a suffix to names of partially complete files in order to avoid clashes from multiple"
+                " clusters working on similar maps (e.g. exporting same map for different timestamps)."
+            ),
+        )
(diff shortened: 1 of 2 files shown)


zigaLuksic:
> I guess looking back at it all now, my main issue with the unique names was that they were all random, not to mention that they weren't cleaned up if something went wrong.
> 
> But not that I'm seeing this suffix thingy and seeing that it's temporary, why bother the user with a temporary suffix, if the end map will not contain this suffix anyway? Could we randomly generate a suffix for a run and append it to all eopatches from this run, and then clean them up? Would this be more elegant? No need to expose it as a parameter then.
> 
> Unless I'm interpreting the "temporary" part of it wrongly.

Well that's the issue. If something goes wrong in the pipeline (don't just think errors inside python, imagine a problem with the cluster) then having a random suffix will result in cleaning-by-hand. This way you wouldn't use a suffix 95% of the time, but have the option to for that 5%.

eogrow/pipelines/export_maps.py
```diff
@@ -74,6 +74,13 @@ class Schema(Pipeline.Schema):
                 " With this parameter you force to always make copies."
             ),
         )
+        temporary_files_suffix: str = Field(
```
mlubej, line 77: this name is the only thing that bothers me in the MR, but I also don't have a better suggestion. The name looks scary, but if you simplify it, we increase risk of misinterpretation.

Perhaps `patch_suffix` or something could also work?

Can you describe the issue again? IIRC you mentioned it was an issue when running for different quarters?

zigaLuksic: `clash_prevention_suffix` :P

The case we had was downloading hella data (mosaicking and all that jazz). So we had to separate the data into quarters for 3 years, or the batch API wouldn't be able to handle it. And then we ingested it as a BYOC.

For each quarter we ran the same pipeline, which extracted the tiff into a map named `BANDS`. So if i tried to run two pipelines in parallel, the exported tiffs had the same names and were overriding eachother

batic: How about a "simple" `additional_suffix` or perhaps more descriptive `distinctive_suffix`?

zigaLuksic: I would like the name to convey that the suffix will NOT be part of the final map. Was also considering `partial_results_suffix`.

batic: How about `interim_results_suffix`

mlubej: > I would like the name to convey that the suffix will NOT be part of the final map.

This is my main issue with this, if it's not a part of the final map, why even bother the user with this in the first place, even if it's just meant for those last 5%.

TBH, I'm fine with any of the choice above (perhaps also `working_suffix`), if this is the only way to go ahead. I was just trying to see if there is a way of automatizing this.

For example, what are the differences between the configs mentioned by @zigaLuksic above? IIRC, it's the variables. Perhaps all available variables could be automatically used in the temporary names? But then again, the used isn't forced to use variables, they could just create multiple configs with different paths, so we're back to square 1

Or perhaps hex-string of the config? it should remain the same for the same configs and would be easy to clean up in case of any failure.

zigaLuksic: using other parameters automatically gives you 'uniqueness of runs', because if it fails something was probably wrong with the config.
We can have one of:
1. auto-unique (enough) suffixes, cannot do cleanup with subsequent runs
2. the user adds a suffix parameter if needed (it has 'no suffix' as default, no reason to set it, usually one only sets around half of params anyway)
3. no suffixes, causing issues with parallelization in temporally-split cases.

mlubej: >  because if it fails something was probably wrong with the config.

Good point, yes.

Then let's go with item 2. I would vote for `interim_suffix` or `working_suffix`. It's different enough to not assume what it is (hopefully), and then you have the description for what it really is.

zigaLuksic: done, also stated that it does not affect results in the first sentence of description

mlubej (approved):

PR #130, opened 2022-11-11
Check compatibility of area/eopatch managers for Batch

zigaLuksic:
It is already checked one way (batch eopatch can only be used with batch area) but it often happens that one uses batch area but not batch eopatch manager. The code fails at the end of the batch download (kinda shitty).

This is a temporary hack that checks the compatibility until eopatch and area managers are reworked.


mlubej (approved):

mlubej:
I like your consistency with tests 👏

PR #131, opened 2022-11-11
Hardcode the compression when saving

zigaLuksic:
The parameter never set to anything other than 1 except by mistake.
With this we get rid of some code complexity and inconsistency.

But i'm not entirely sure if this is a step in the right direction :/

### DIFF
--- a/eogrow/pipelines/download.py
+++ b/eogrow/pipelines/download.py
@@ -79,7 +79,6 @@ class BaseDownloadPipeline(Pipeline, metaclass=abc.ABCMeta):
             description="Storage manager key pointing to the path where downloaded EOPatches will be saved."
         )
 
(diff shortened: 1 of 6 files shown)


mlubej:
I thought the idea was to convert the parameter to a boolean one, which is True (to use compression 1) by default, otherwise the user can specify they are not interested in compression, since they want to prioritize speed and not disk space.

zigaLuksic:
> I thought the idea was to convert the parameter to a boolean one, which is True (to use compression 1) by default, otherwise the user can specify they are not interested in compression, since they want to prioritize speed and not disk space.
> 
> Would that still make sense, or is it too similar to the situation before?

The idea was to get rid of a parameter, because we often have a lot of parameters we never touch

zigaLuksic:
not doing it (for now)

PR #132, opened 2022-11-16
Remove utils.enum due to non-use

zigaLuksic:


zigaLuksic:
wrong base of branch, will fix and reopen new one

PR #138, opened 2022-11-24
Remove regions from area manager

zigaLuksic:
Regions were a feature that was never used since LCMS changed to eo-grow. They needlessly complicate the base area manager interface, especially since child area managers might not even support it!!

It is thus removed, if anyone wants to do stuff with regions... make a custom upgrade to your choice of area manager or create EOPatch lists for different regions.


zigaLuksic:
Some issues in our test suite came to life with this, so I fixed the underlying error (which was that the tests were stupid and did not test what the thought they were testing)

eogrow/core/area/base.py
```diff
@@ -189,16 +170,10 @@ def _save_area_geometry(self, area_geometry: Geometry, filename: str) -> None:
                 local_file.path, driver="GPKG", encoding="utf-8", engine=self.storage.config.geopandas_backend
             )
 
-    def _create_and_save_grid(self, grid_filename: str, ignore_region_filter: bool) -> List[gpd.GeoDataFrame]:
+    def _create_and_save_grid(self, grid_filename: str) -> List[gpd.GeoDataFrame]:
```
jgersak, line 173:
```suggestion
    def _create_and_save_grid(self,*, grid_filename: str) -> List[gpd.GeoDataFrame]:
```

eogrow/core/area/base.py
```diff
@@ -57,71 +53,60 @@ def __init__(self, config: Schema, storage: StorageManager):
...
-            grid = self._create_and_save_grid(grid_filename, ignore_region_filter)
+            grid = self._create_and_save_grid(grid_filename)
 
         if add_bbox_column:
             self._add_bbox_column(grid)
 
         return grid
 
-    def cache_grid(self, grid: Optional[List[gpd.GeoDataFrame]] = None, ignore_region_filter: bool = False) -> None:
+    def cache_grid(self, grid: Optional[List[gpd.GeoDataFrame]] = None) -> None:
```
jgersak, line 100:
```suggestion
    def cache_grid(self, *, grid: Optional[List[gpd.GeoDataFrame]] = None) -> None:
```

PR #141, opened 2022-11-28
ZipMapPipeline

zigaLuksic:
Adds a new ZipMapPipeline and deprecates the MappingPipeline since it is subsumed by it's successor.

The MappingPipeline is not yet removed, it just emits a warning on use, but is no longer part of the test suite.


eogrow/pipelines/zipmap.py
```diff
@@ -0,0 +1,133 @@
...
+        load_nodes = self.get_load_nodes()
+        merge_node = EONode(MergeEOPatchesTask(), inputs=load_nodes)
+
+        mapping_node = self.get_zipmap_node(merge_node)
+
+        save_path = self.storage.get_folder(self.config.output_folder_key, full_path=True)
+        save_task = SaveTask(
+            save_path,
+            config=self.sh_config,
+            features=[self.config.output_feature, FeatureType.BBOX, FeatureType.TIMESTAMP],
```
veseln, line 127: Will this work if the bbox/timestamp are not loaded on input?

eogrow/pipelines/zipmap.py
```diff
@@ -0,0 +1,133 @@
...
+from ..core.schemas import BaseSchema
+from ..utils.filter import get_patches_with_missing_features
+from ..utils.meta import import_object
+from ..utils.types import Feature, FeatureSpec
+
+LOGGER = logging.getLogger(__name__)
+
+
+class InputFeatureSchema(BaseSchema):
+    feature: Feature = Field(description="Which features to load from folder.")
```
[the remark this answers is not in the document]
zigaLuksic: you are correct that for the loading step that would be the superior choice. In fact we even started out with that idea :D

But the issue here is that once the data is zipped together, you need to know the order of the features in the data sandwich.
In the current way, the index of the feature coincides with the index of the data. In the nested structures we'd need to adapt a convention of how to translate the feature specification to the indices.

With @mlubej we agreed that the list approach (although more verbose) might help avoid issues. What do you think?

eogrow/pipelines/zipmap.py
zigaLuksic: yes, I also had to test this cause I didn't remember
for saving this parameter serves as a filter, and not precise specification

eogrow/pipelines/zipmap.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_load_nodes(self) -> List[EONode]:
+        """Prepare all nodes with load tasks."""
+        load_schema: DefaultDict[str, Set[FeatureSpec]] = defaultdict(set)
+        for input_feature in self.config.input_features:
+            features_to_load = load_schema[input_feature.folder_key]
+            features_to_load.add(input_feature.feature)
+
+            if input_feature.include_bbox_and_timestamp:
+                features_to_load.add(FeatureType.BBOX)
```
veseln, line 93: What is the main reason that we are handling the bbox and timestamps separately? 


 It does add quite some complexity to the code so I want to understand the motivations behind it. Is it just to not load data that we don't need? Or is there another reason?

zigaLuksic: Unlike `SaveTask`, loading demands the features exist.
So ofcourse if the features are not temporal, it makes no sense to load the timestamps (a lot less IO calls). But in some cases when we have storages with lots of subfolders, some don't contain a BBox, and this was meant to avoid such things.

We again weren't sure whether this should be always on, or just on by default.

veseln: I understand. I think you are correct, let's leave it like it is.

veseln:
I have no further comments. Seems okay to me, will have further comments once I try to use it for a few times and things don't work as I imagined 🤣

zigaLuksic:
> I have no further comments. Seems okay to me, will have further comments once I try to use it for a few times and things don't work as I imagined rofl

I wouldn't want it any other way :P it's the only way to improve the tools

PR #145, opened 2022-12-02
Switch grids alternative pipeline

zigaLuksic:
As discussed I implemented a simpler version for sub-splitting grids.
It is now area manager agnostic (only needs grid) and produces output manager, at the cost of some minor things being locked in place.

The test files are huge, i know :( ideas welcome


eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+
+class SplitGridPipeline(Pipeline):
+    """Pipeline that creates a finer grid and splits EOPatches accordingly.
+
+    The new grid is output as a geopackage file, which can be used with a `CustomAreaManager`.
+    """
+
+    class Schema(Pipeline.Schema):
+        input_folder_key: str = Field(
+            description="A storage manager key pointing to the folder from where data will be loaded."
```
mlubej, line 34: "A storage manager key pointing to the folder where the data will be loaded."*

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        eopatch_output_folder_key: str = Field(
+            description="A storage manager key pointing to the folder to where data will be saved."
```
mlubej, line 37: "A storage manager key pointing to the folder where the data will be saved."*

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        grid_output_folder_key: str = Field(
+            description="A storage manager key of where to save the resulting split grid."
+        )
+        subsplit_grid_filename: str = Field(
+            description="Filename of new grid, which can be used in `CustomAreaManager`.", regex=r"^.+\.gpkg$"
+        )
+        features: List[Feature] = Field(description="Which features will be loaded and joined.")
```
mlubej, line 45: What is meant by "joined" here?

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+        raise_misaligned: bool = Field(
+            True,
+            description=(
+                "If True this will raise an error if splitting or joining any spatial raster EOPatch feature would "
+                "cause a misalignment. If False, misalignment issues will be ignored."
```
mlubej, line 50: probably can be simplified a bit, perhaps:

"Raise an error if spatially splitting the EOPatch produces misalignments."

I also removed the "or joining" part since this is probably split only?

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+
+        split_x: int = Field(2, description="Number of splits done on each EOPatch along the x-axis.")
+        split_y: int = Field(2, description="Number of splits done on each EOPatch along the y-axis.")
+
+        buffer: Union[Literal["auto"], Tuple[float, float]] = Field(
+            "auto",
+            description=(
+                "How large is the original EOPatch buffer, provided as `(buffer_x, buffer_y)` in CRS units. Applies"
+                " same buffer to split EOPatches. The pipeline tries to obtain this information from area manager, but"
+                " not all area manager classes are supported."
```
mlubej, line 62: Replace first sentence with:  "Amount of buffer present on existing EOPatches, provided as `(buffer_x, buffer_y)` in CRS units."

>  The pipeline tries to obtain this information from area manager, but not all area manager classes are supported.

What if the new process saves the buffer as a column in the geopackage? would this provide any value for easier interpretation in such cases?

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+            ),
+        )
+        # TODO: could add AOI file?
```
mlubej, line 65: Please explain this, not sure I understand how this would work.

eogrow/pipelines/split_grid.py
```diff
@@ -0,0 +1,184 @@
...
+
+        return exec_args
+
+    def _get_features(self) -> List[FeatureSpec]:
+        """Provides features that will be transformed by the pipeline."""
+        meta_features = [FeatureType.BBOX]
+        if any(f_type.is_temporal() for f_type, _ in self.config.features):
+            meta_features += [FeatureType.TIMESTAMP]
+
+        return self.config.features + meta_features
```
mlubej, line 160: This seems quite generic, perhaps could go into utils or something?

eogrow/utils/grid.py
```diff
@@ -95,3 +95,35 @@ def get_enclosing_bbox(bboxes: Sequence[BBox]) -> BBox:
...
+    """Splits a BBox into multiple smaller BBoxes with new names generated for them.
+
+    The `buffer` parameters describe the buffer of the original BBox, which is copied to the split ones.
+    """
+    name, bbox = named_bbox
+    min_x, min_y = bbox.lower_left
+    max_x, max_y = bbox.upper_right
+
+    ll_xs, x_step = np.linspace(min_x + buffer_x, max_x - buffer_x, split_x, endpoint=False, retstep=True)
+    ll_ys, y_step = np.linspace(min_y + buffer_y, max_y - buffer_y, split_y, endpoint=False, retstep=True)
```
mlubej, line 117: `ll_xs` and `ll_ys` not descriptive enough.

tests/test_config_files/switch_grids/new/switch_utm.json
```diff
@@ -0,0 +1,33 @@
...
+    "area": {
+      "manager": "eogrow.core.area.UtmZoneAreaManager",
+      "area_filename": "test_area.geojson",
+      "area_buffer": 0.01,
+      "patch_size_x": 3600,
+      "patch_size_y": 3000,
+      "patch_buffer_x": 120,
+      "patch_buffer_y": 50,
+      "offset_x": 100,
+      "offset_y": 20
```
mlubej, line 16: do these offsets not affect the split? I guess not, because the splitter is working directly with produced geometries, but just wanted to make sure.

mlubej (commented):
There is one empty file that was added: `tests/test_config_files/switch_grids/new/resulting_area_manager.json`

eogrow/pipelines/split_grid.py
zigaLuksic: Done, added "from" to the end

eogrow/pipelines/split_grid.py
zigaLuksic: done

eogrow/pipelines/split_grid.py
zigaLuksic: Old docstring, changed to "Which features will be loaded and adapted to new grid."

eogrow/pipelines/split_grid.py
zigaLuksic: done

eogrow/pipelines/split_grid.py
zigaLuksic: Changed docstring

Since all EOPatches have the same buffer, it would be quite a bit of memory for little gain. Also remains difficult for `CustomAreaManager`, where patches can have distinct buffer values in the column, which is also odd. So not sure.

tests/test_config_files/switch_grids/new/switch_utm.json
zigaLuksic: no, it works entirely on geometries
i wanted to leave them in to ensure it doesnt inerfere

eogrow/utils/grid.py
zigaLuksic: Would it suffice if i add a comment `# the ll_ prefix stands for lower_left`?

eogrow/pipelines/split_grid.py
zigaLuksic: i'll open an issue with this suggestion

eogrow/pipelines/split_grid.py
zigaLuksic: If an AOI file is provided, all EOPatches that do not lie inside the area are discarded.
This was done by previous implementation, since Batch and UTM area managers carry an AOI.

mlubej: Interesting use case, one could use this pipeline without splitting further and just using the input AOI as a filter, e.g. for running several per-region experiments.

But I don't see how this functionality would be useful when doing a split, not how the split functionality would be useful when doing the filtering, so IDK.

eogrow/utils/grid.py
mlubej: Perhaps there is no need to use `ll` in the first place and would be more intuitive with:

`x_edges, x_step = ...`

eogrow/pipelines/split_grid.py
zigaLuksic: Let's see if i can ASCII-art myself out of this...

Assume this is the AOI and is covered by a single EOPatch in the original grid.
```
┌──────┐
│ ▉    │
│ ▉▉   │
│ ▉▉▉▉ │
│  ▉▉  │
└──────┘
```
By splitting into 2x2 we currently get 
```
┌───┐┌───┐
│ ▉ ││   │
│ ▉▉││   │
└───┘└───┘
┌───┐┌───┐
│ ▉▉││▉▉ │
│  ▉││▉  │
└───┘└───┘
```
but there is no need for the upper right EOPatch. However in order for the pipeline to know that, it would need to have access to the AOI. One option is to take the one from the AreaManager, but it'd say its better it's an explicit input.

mlubej: I think it's good if such an input functionality is added, but would set the default to the AOI from AreaManager

mlubej: nice art!

zigaLuksic: What if the default is that no such filtration is used?

mlubej: I assumed that _original AOI_ == _no filtration_, or did you mean that also the upper right split eopatch is retained in this case?

zigaLuksic: If no AOI is provided, all the resulting EOPatches would be retained. If an AOI is provided, only EOPatches intersecting it would be retained.

mlubej: From the point of view of not making decisions on behalf of the user, I completely agree with this.

However, from the other point of view, we might be pushing people to use unnecessary resources for processing split eopatches which fall out of the AOI they're originally interested in.

idk.. should we vote? I would still vote for the latter, because if the user would provide the finer grid, these extra eopatches wouldn't be present.

mlubej: This can be resolved, right? Due to https://github.com/sentinel-hub/eo-grow/pull/150

PR #148, opened 2022-12-02
Define new BaseAreaManager and create custom grid version of it

zigaLuksic:
In order to keep chunks smaller, i'll add them in such small updated.

For now I focused on supporting the bare minimum, and things will only be added if there is no better way of doing it.
I decided to transition grids from lists to dictionaries, because it is now impossible to have two grid files for the same CRS.

the biggest issue I have is how to name the column with eopatch names in the grid representation. I was considering just fixing it to `eopatch_name` so that the rest of the code does not need to use `self.area_manager.NAMES_COLUMN`

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -1,7 +1,8 @@
 """Implementation of the base AreaManager."""
 import logging
+from abc import ABCMeta, abstractmethod
 from collections import defaultdict
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 import fiona
 import fs
@@ -248,3 +249,92 @@ class AreaManager(EOGrowObject):
         for bbox_df in grid:
             crs = CRS(bbox_df.crs)
             bbox_df["BBOX"] = bbox_df.geometry.apply(lambda geo: BBox(geo.bounds, crs))  # noqa B023
+
+
+class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
+    """A manager for the AOI and how it is split into EOPatches"""
+
+    NAME_COLUMN = "eopatch_name"
+
+    class Schema(ManagerSchema):
+        ...
+
+    config: Schema
+
+    def __init__(self, config: Schema, storage: StorageManager):
+        """
+        :param config: A configuration file
+        :param storage: An instance of StorageManager class
+        """
+        super().__init__(config)
+
+        self.storage = storage
+
+    @abstractmethod
+    def get_area_geometry(self, *, crs: CRS = CRS.WGS84) -> Geometry:
+        """Provides a single geometry object of entire AOI"""
+
+    def get_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
+        """Provides a grid of bounding boxes which divide the AOI. Uses caching to avoid recalculations.
+
+        The grid is split into different CRS zones.
+
+        The `geometry` column serves as BBox geometry definition. The `geom.bounds` is taken as the definition.
+
+        The EOPatch names are stored in a column with identifier `self.NAME_COLUMN`.
+
+        :return: A dictionary of GeoDataFrames that defines how the area is split into EOPatches.
+        """
+        grid_path = fs.path.combine(self.storage.get_cache_folder(), self.get_grid_cache_filename())
+
+        if self.storage.filesystem.exists(grid_path):
+            return self._load_grid(grid_path)
+
+        grid = self._create_grid()
+        self._save_grid(grid, grid_path)
+        return grid
+
+    @abstractmethod
+    def _create_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
+        """Defines a new grid, which encodes how the area is split into EOPatches.
+
+        The grid is split into different CRS zones.
+
+        The `geometry` column serves as BBox geometry definition, with the implication that for a geometry `geom`
+        the BBox is defined as `geom.bounds`.
+
+        The EOPatch names are stored in a column with identifier `eopatch_name`.
+        """
+
+    def _load_grid(self, grid_path: str) -> Dict[CRS, gpd.GeoDataFrame]:
+        """A method that loads bounding box grid saved in a cache folder"""
+        LOGGER.info("Loading grid from %s", grid_path)
+
+        grid = {}
+        with LocalFile(grid_path, mode="r", filesystem=self.storage.filesystem) as local_file:
+            for crs_layer in fiona.listlayers(local_file.path):
+                data = gpd.read_file(local_file.path, layer=crs_layer, engine=self.storage.config.geopandas_backend)
+                grid[CRS(data.crs)] = data
+
+        return grid
+
+    def _save_grid(self, grid: Dict[CRS, gpd.GeoDataFrame], grid_path: str) -> None:
+        """A method that saves bounding box grid in a cache folder"""
+        LOGGER.info("Saving grid to %s", grid_path)
+
+        with LocalFile(grid_path, mode="w", filesystem=self.storage.filesystem) as local_file:
+            for _, crs_grid in grid.items():
+                crs_grid.to_file(
+                    local_file.path,
+                    driver="GPKG",
+                    encoding="utf-8",
+                    layer=f"Grid EPSG:{crs_grid.crs.to_epsg()}",
+                    engine=self.storage.config.geopandas_backend,
+                )
+
+    @abstractmethod
+    def get_grid_cache_filename(self) -> str:
+        """Provides a filename that is used for caching the grid, including the file extensions (likely .gpkg).
+
+        Should ensure that two different grids don't clash.
+        """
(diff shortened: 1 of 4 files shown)


eogrow/core/area/base.py
```diff
@@ -248,3 +249,92 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
...
+        self.storage = storage
+
+    @abstractmethod
+    def get_area_geometry(self, *, crs: CRS = CRS.WGS84) -> Geometry:
+        """Provides a single geometry object of entire AOI"""
+
+    def get_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
+        """Provides a grid of bounding boxes which divide the AOI. Uses caching to avoid recalculations.
+
+        The grid is split into different CRS zones.
```
mlubej, line 280:
```suggestion
        The grid is split into different UTM CRS zones.
```

eogrow/core/area/base.py
```diff
@@ -248,3 +249,92 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
...
+
+        The `geometry` column serves as BBox geometry definition. The `geom.bounds` is taken as the definition.
```
mlubej, line 282: Funny sentence:

```suggestion
        The `bounds` property of the BBox geometries is taken as definition.
```

eogrow/core/area/base.py
```diff
@@ -248,3 +249,92 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
...
+    def _create_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
+        """Defines a new grid, which encodes how the area is split into EOPatches.
+
+        The grid is split into different CRS zones.
+
+        The `geometry` column serves as BBox geometry definition, with the implication that for a geometry `geom`
+        the BBox is defined as `geom.bounds`.
+
+        The EOPatch names are stored in a column with identifier `eopatch_name`.
+        """
```
mlubej, line 307: Not sure what happened here, function has almost the same docstring and nothing else (no content, no return)

Perhaps you're setting things up in advance?

eogrow/core/area/base.py
```diff
@@ -248,3 +249,92 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
...
+
+    def _load_grid(self, grid_path: str) -> Dict[CRS, gpd.GeoDataFrame]:
+        """A method that loads bounding box grid saved in a cache folder"""
+        LOGGER.info("Loading grid from %s", grid_path)
+
+        grid = {}
+        with LocalFile(grid_path, mode="r", filesystem=self.storage.filesystem) as local_file:
+            for crs_layer in fiona.listlayers(local_file.path):
+                data = gpd.read_file(local_file.path, layer=crs_layer, engine=self.storage.config.geopandas_backend)
+                grid[CRS(data.crs)] = data
```
mlubej, line 317: What if the loaded CRSs arent in UTM, is that a problem?

eogrow/core/area/base.py
```diff
@@ -248,3 +249,92 @@ def _add_bbox_column(grid: List[gpd.GeoDataFrame]) -> None:
...
+                    layer=f"Grid EPSG:{crs_grid.crs.to_epsg()}",
+                    engine=self.storage.config.geopandas_backend,
+                )
+
+    @abstractmethod
+    def get_grid_cache_filename(self) -> str:
+        """Provides a filename that is used for caching the grid, including the file extensions (likely .gpkg).
+
+        Should ensure that two different grids don't clash.
+        """
```
mlubej, line 340: I'm assuming you're setting things up ahead here as well

mlubej (commented):
Review part I

eogrow/core/area/base.py
zigaLuksic: sadly, wishful thinking :)
i guess it's my fault for saying CRS zones instead of CRS parts/layers/whatever, but as we talked, we cannot afford to restrict ourselves to UTM zones only.

eogrow/core/area/base.py
zigaLuksic: we are not UTM restricted

eogrow/core/area/base.py
zigaLuksic: went with "The `bounds` properties of the geometries are taken as BBox definitions."

eogrow/core/area/base.py
mlubej: Can you give an example how these parts are defined for other CRS cases? I thought this is only done for UTM CRS, because we have utility functions which detects the native UTM CRS of the EOPatch and it know in which part to put it.

eogrow/core/area/base.py
mlubej: Didn't understand completely the abstract classes. Learned something new. Resolving

eogrow/core/area/base.py
mlubej: This question goes together with the topic above, where it was not clear to me how CRS parts are defined for non-UTM CRS cases.

eogrow/core/area/base.py
zigaLuksic: The responsibility of each manager is to provide a collection of geometries, each with it's respective CRS. From these geometries + CRS the BBoxes are then created.

In the case of `UtmZoneAreaManager` and `BatchAreaManager` the CRS will be UTM CRS. In the case of `CustomGridAreaManager` these will be whatever the geopackage wants. As discussed on the meeting we are trying to avoid streamlining everything to CRS. An example was the project done in epsg 3035 (if i recall correctly) which had reference data given in the mentioned CRS. Reprojecting to UTM CRS would produce warping artefacts as well as 'no-data' areas.

eogrow/core/area/custom_grid.py
```diff
@@ -81,3 +81,46 @@ def _construct_file_path(self, *, prefix: str, suffix: str = "gpkg", **_: Any) -
...
+
+        return grid
+
+    def get_area_geometry(self, *, crs: CRS = CRS.WGS84) -> Geometry:
+        all_grid_gdfs = list(self.get_grid().values())
+        area_df = concat_gdf(all_grid_gdfs, reproject_crs=crs)
+
+        LOGGER.info("Calculating a unary union of the area geometries")
+        area_shape = shapely.ops.unary_union(area_df.geometry)
+        return Geometry(area_shape, crs)
```
zigaLuksic, line 120: @mlubej  now that i think about this, the 'area geometry' is usually the buffered and simplified AOI, but in the case of the custom grid I just took all the elements of the grid. Do you think it would be more correct to take the union of all *geometries* (and not the bboxes that they define) of the custom grid? Usually a custom grid will contain preciselly BBox geometries, but not necessarily.

mlubej: If the custom grid is pre-defined, do you even have access to the original geometries? I imagine if the user just wants to proved the new grid, they are not interested in the original geometries anymore (maybe)

eogrow/core/area/custom_grid.py
```diff
@@ -81,3 +81,46 @@ def _construct_file_path(self, *, prefix: str, suffix: str = "gpkg", **_: Any) -
...
+class NewCustomGridAreaManager(BaseAreaManager):
+    """Area manager that works with a pre-defined grid of EOPatches"""
+
+    class Schema(BaseAreaManager.Schema):
+        grid_folder_key: str = Field("input_data", description="Which folder the grid file is in.")
+        grid_filename: Path = Field(
+            description=(
+                "A Geopackage with a collection of bounding boxes and attributes that will define EOPatches. In"
+                " case bounding boxes are in multiple CRS then each Geopackage layer should contain bounding boxes"
+                " from one CRS."
```
mlubej, line 95: Just checking, is it really possible to put layers with different CRS to the same geopackage?

eogrow/core/area/custom_grid.py
```diff
@@ -81,3 +81,46 @@ def _construct_file_path(self, *, prefix: str, suffix: str = "gpkg", **_: Any) -
...
+            ),
+            regex=r"^.+\..+$",
+        )
+        name_column: str = Field(description="Name of the column containing EOPatch names.")
+
+    config: Schema
+
+    def _create_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
```
mlubej, line 103: Perhaps this functions name is misleading, since the grid already exists in this case. 

Maybe `_create_grid_dict` or `_create_grid_map` ? (if yes, fix also in the base class)

eogrow/core/area/custom_grid.py
zigaLuksic: I actually think it already does what i wanted it to do, so i'm resolving this :see_no_evil:

eogrow/core/area/custom_grid.py
zigaLuksic: I consider a `grid` to be an object of type `Dict[CRS, gpd.GeoDataFrame]`. And this method has to provide such a grid, given the input parameters. So here it does create such a grid, and it does so by loading a geopackage.

Would the name `_provide_grid` or `_prepare_grid` be more sensible in your opinion? The names kind of work for splitter-based classes and the batch one as well.

mlubej: you know what, it's actually fine, I had a wrong mapping of names in my head. my bad

mlubej (approved):

PR #150, opened 2022-12-07
Add AOI filtration option to switch grids pipeline

zigaLuksic:
A proposal for the feature we discussed.

In the end I agree with you that it should be set to the AreaManagers AOI by default. If the AOI file would be used, it might clash with the AOI (since area manager sometimes simplify or buffer the geometry from the file).

Some bloat comes from trying to avoid reprojecting the area too many times.


eogrow/pipelines/split_grid.py
```diff
@@ -59,7 +61,14 @@ class Schema(Pipeline.Schema):
                 " not all area manager classes are supported."
             ),
         )
-        # TODO: could add AOI file?
+
+        keep_aoi_only: bool = Field(
```
mlubej, line 65: Just to understand better: in https://github.com/sentinel-hub/eo-grow/pull/145#discussion_r1038185372 we were discussing if the user could provide an AOI for some further filtration - I thought you meant this for any generic AOI, meaning that the user could also provide a sub-part of the original AOI, but IIRC now the user only has the option to keep or discard eopatches which would fall out of the AM-AOI after splitting?

zigaLuksic: yes, we were discussing many options. Technically the manager-based filtration is always desired. The custom AOI filtration not that often. Since we're outputting a gpkg, the user can always modify it to a sub-area by hand if desired (some redundant data needs to be deleted as well)

Or did you get too excited about that, and now I need to add it? :P

eogrow/pipelines/split_grid.py
```diff
@@ -139,10 +168,12 @@ def get_execution_arguments(  # type: ignore[override]
         exec_args: List[Dict[EONode, Dict[str, object]]] = []
         for (orig_name, _), split_bboxes in bbox_splits:
             single_exec: Dict[EONode, Dict[str, object]] = {load_node: dict(eopatch_folder=orig_name)}
+            # Since some bboxes might get filtered out, the remaining slice and save nodes should get None arguments
+            split_bboxes_iter = it.chain(split_bboxes, it.repeat((None, None)))
 
-            for i, (subbox_name, subbox) in enumerate(split_bboxes):
-                single_exec[slice_nodes[i]] = dict(bbox=subbox)
-                single_exec[save_nodes[i]] = dict(eopatch_folder=subbox_name)
+            for slice_node, save_node, (subbox_name, subbox) in zip(slice_nodes, save_nodes, split_bboxes_iter):
```
mlubej, line 174: I'm not sure if I understand how the order of the `slice_nodes` corresponds to the order of `split_bboxes_iter`, can you explain?

Let's assume we have a (2,2) split, and after the split any of the subpatches (up to 3) could fall out of the AM AOI, while in `split_bboxes_iter` the `(None, None`) is appended to the end, so this is where it breaks my brain. What am I misunderstanding?

zigaLuksic: Slice nodes, or rather slice branches (slice + save) are order agnostic. It is important here that slice nodes and save nodes are correctly ordered together, since they are passed the same BBox. Other than that it does not matter which branch gets which BBox.

Both tasks also have a 'fast-forward' option when the BBox is set to None. For slice this means that only non-spatial data is copied over and the EOPatch is shipped (we avoid all computation aside from some pointer shuffling, good enough), and SaveTask doesn't save if the name is None.

So if we have 4 subboxes, and only 3 get saved, the filtration will give us only 1 named bbox. But we have 4 slice branches (since we know in advance what the upper limit is and create that many). So the remaining three branches get None arguments so that they do nothing.

mlubej: Ah, OK, I understand, the slice node just performs the slice, depending on the input, and the input (the bbox) defines how the original eopatch will be sliced. Now it's so clear that I don't know what I was thinking.

mlubej (approved):

PR #151, opened 2022-12-07
Rework utm zone area manager

zigaLuksic:


eogrow/core/area/utm.py
```diff
@@ -123,3 +131,105 @@ def _check_target_compatibility(self, target_area_manager: AreaManager) -> None:
...
+
+        if self.config.area_of_interest.buffer is not None:
+            area_shape = area_shape.buffer(self.config.area_of_interest.buffer)
+
+        if self.config.area_of_interest.simplification_factor is not None:
+            area_shape = area_shape.simplify(self.config.area_of_interest.simplification_factor, preserve_topology=True)
+            LOGGER.info("Simplified area shape has %d points", count_points(area_shape))
+
+        LOGGER.info("Finished processing AOI geometry")
+        return Geometry(area_shape, CRS(area_df.crs)).transform(crs)
```
mlubej, line 218: Why is the geometry transformed rather than kept in the original crs? Does it need to be WGS in order for the `UtmZoneSplitter` to work?

zigaLuksic: my line of thought was that I don't want certain pipelines to fail because the user provided the geometry in a weird CRS two weeks ago and forgot about it. Either we request a specific CRS which we need, or a default one is used.

I am willing to change my mind if you have any good arguments and/or options

mlubej (approved):

PR #152, opened 2022-12-12
Batch area manager rework

zigaLuksic:
1. Extracts dealing with an AOI into a `BaseSplitterAreaManager` (since it's common to both splitter based managers)
2. Implements new Batch area manager
3. adds tests for it that (with some mocking)
4. I learned some new tricks so I fixed some old tests as well before i forgot about it

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -254,9 +254,6 @@ class AreaManager(EOGrowObject):
 class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
     """A manager for the AOI and how it is split into EOPatches"""
 
-    # NOTE: since the BaseAreaManager is an abstract class, some core functionalities (caching etc.) are instead
-    #   tested in the test suite of the UtmZoneAreaManager.
-
     NAME_COLUMN = "eopatch_name"
 
     class Schema(ManagerSchema):
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -2,7 +2,7 @@
 import logging
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
-from typing import Dict, List, Optional
+from typing import Dict, List, Literal, Optional
 
 import fiona
 import fs
@@ -21,6 +21,21 @@ from ..schemas import BaseSchema, ManagerSchema
 from ..storage import StorageManager
 
 LOGGER = logging.getLogger(__name__)
+
+
+class AreaSchema(BaseSchema):
+    filename: str
+    buffer: Optional[float] = Field(
+        description=(
+            "Buffer that will be applied to AOI geometry. Buffer has to be in the same units as AOI CRS. "
+            "In case buffer is too small, relatively to AOI size, it won't have any effect."
+        ),
+    )
+    simplification_factor: Optional[float] = Field(
+        description=(
+            "A tolerance factor in CRS units how much the buffered area geometry will be simplified before splitting."
+        ),
+    )
 
 
 class AreaManager(EOGrowObject):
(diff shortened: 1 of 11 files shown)


eogrow/core/area/base.py
```diff
@@ -334,3 +331,46 @@ def get_grid_cache_filename(self) -> str:
...
+
+        if self.config.area.buffer is not None:
+            area_shape = area_shape.buffer(self.config.area.buffer)
+
+        if self.config.area.simplification_factor is not None:
+            area_shape = area_shape.simplify(self.config.area.simplification_factor, preserve_topology=True)
+            LOGGER.info("Simplified area shape has %d points", count_points(area_shape))
+
+        LOGGER.info("Finished processing AOI geometry")
+        return Geometry(area_shape, CRS(area_df.crs)).transform(crs)
```
mlubej, line 376: In what sense is this part not a generic piece of code that it shouldn't be in `BaseAreaManager`?

zigaLuksic: Because it only works for splitter based approaches, which is why it's in `BaseSplitterAreaManager` :P
Joking aside, this was actually THE method that broke type-safety previously, because `CustomGridAreaManager` does not have an AOI file, just a grid. Hence this method was broken for it.

tests/test_core/test_area/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+    def _create_grid(self) -> Dict[CRS, GeoDataFrame]:
+        bboxes = [BBox((0, 0, 1, 1), CRS.WGS84), BBox((1, 1, 2, 2), CRS.WGS84), BBox((0, 0, 1, 1), CRS(3035))]
+        return {
+            CRS.WGS84: GeoDataFrame(
+                data={"eopatch_name": ["beep", "boop"]},
+                geometry=[bbox.geometry for bbox in bboxes[:2]],
+                crs=CRS.WGS84.pyproj_crs(),
+            ),
+            CRS(3035): GeoDataFrame(
+                data={"eopatch_name": ["bap"]}, geometry=[bboxes[2].geometry], crs=CRS(3035).pyproj_crs()
```
mlubej, line 27: next time please use scat singing sounds as names

tests/test_core/test_area/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_get_grid_caching(storage):
+    """Tests that subsequent calls use the cached version of the grid. NOTE: implementation specific test!"""
+    manager = DummyAreaManager.from_raw_config({}, storage)
+
+    grid1 = manager.get_grid()
+    # the second call should not create a new split
+    with patch.object(DummyAreaManager, "_create_grid") as creation_mock:
+        grid2 = manager.get_grid()
+        creation_mock.assert_not_called()
```
mlubej, line 46: cool

tests/test_core/test_area/test_utm.py
```diff
@@ -72,7 +72,7 @@ def test_get_grid_caching(storage, config):
 
     grid1 = area_manager.get_grid()
     # the second call should not create a new split
-    with mock.patch.object(UtmZoneAreaManager, "_create_new_split") as creation_mock:
+    with patch.object(UtmZoneAreaManager, "_create_new_split") as creation_mock:
```
mlubej, line 75: Is it necessary to test each `AreaManager` separately, or could you just test the `BaseAreaManager` functionality for caching?

mlubej (approved):
The changes make sense to me, the additions are harder to review, but the logic is solid and the issues are less likely because we're improving existing code.

eogrow/core/area/base.py
mlubej: Discussed in private, a direction for using a util function will be pursued.

zigaLuksic:
> I prefer it more now than before +1

<3

PR #154, opened 2022-12-16
Add eopatch manager functionality to area manager

zigaLuksic:
adds patch list functionality as part of the area manager

area managers can now output named bboxes, which is what eopatch managers used to do. Feels crazy that it seems to only be 20 LOC, maybe i'm overlooking something or it shows that the reowork uses a much better abstraction

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -2,7 +2,7 @@
 import logging
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
-from typing import Dict, Iterable, List, Literal, Optional, Tuple
+from typing import Dict, List, Literal, Optional, Tuple
 
 import fiona
 import fs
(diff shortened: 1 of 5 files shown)


eogrow/core/area/base.py
```diff
@@ -38,6 +39,11 @@ class AreaSchema(BaseSchema):
     )
 
 
+class PatchListSchema(BaseSchema):
+    input_folder_key: str = Field(description="The storage manager key pointing to the folder containing the file.")
+    filename: str = Field(description="A JSON file containing a list of EOPatch names.", regex=r"^.+\.(json|JSON)$")
```
mlubej, line 44: if you want to get regex fancy, you can also use the case insensitive flag

```suggestion
    filename: str = Field(description="A JSON file containing a list of EOPatch names.", regex=r"^.+\.(?i:json)$")
```

mlubej (approved):

mlubej (approved):
Nothing out of the ordinary, pretty straight forward changes. 👍

eogrow/core/area/base.py
zigaLuksic: it does work, but as discussed privately, we'll keep the current one so that the regex is a bit easier to read

PR #155, opened 2022-12-19
Add temporary fix to overview resampling

zigaLuksic:


eogrow/utils/map.py
```diff
@@ -124,6 +125,14 @@ def cogify(
     if dtype is not None:
         gdaltranslate_options += f" -ot {GDAL_DTYPE_SETTINGS[dtype]}"
 
+    if version < "3.6.0" and resampling == "MODE":
+        warnings.warn(
+            "GDAL versions below 3.6.0 have issues with `MODE` overview resampling. Trying to fix issue by setting"
+            " GDAL_OVR_CHUNK_MAX_SIZE to a large integer.",
```
mlubej, line 131: Perhaps just to be more explicit?

```suggestion
            " GDAL_OVR_CHUNK_MAX_SIZE to a large integer (2100000000).",
```

zigaLuksic: was thinking about it before but thought "if they need to know the value, the warning points them to the code". I can add it to the warning itself if you fancy

PR #158, opened 2022-12-23
The great switch

zigaLuksic:
Ported all the pipelines to the new managers.
Removed only what needed to be removed from the old ones (utm_area and eopatch were causing tests to fail, so they were removed already) and the rest will be removed later.

Went surprisingly smoothly. We really managed to follow:
1. Make the change easy (this is hard)
2. Make the easy change


eogrow/pipelines/merge_samples.py
```diff
@@ -29,7 +29,12 @@ class Schema(Pipeline.Schema):
             description="Dictionary of all features for which samples are to be merged."
         )
         include_timestamp: bool = Field(False, description="Whether to also prepare an array of merged timestamps.")
-        id_filename: Optional[str] = Field(description="Filename of array holding patch id of concatenated features")
+        id_filename: Optional[str] = Field(
+            description=(
+                "Filename of array holding patch ID of concatenated features. The patch ID is the index of the patch in"
+                " the final patch list, any filtration of the patch list will impact the results."
```
mlubej, line 35: Is it now any harder to obtain the eopatch of interest if the index of the patch ID is affected by the filtration?

eogrow/pipelines/testing.py
```diff
@@ -47,11 +47,11 @@ def run_procedure(self) -> Tuple[List, List]:
 
         self.area_manager.get_area_geometry()
         grid = self.area_manager.get_grid()
-        num_patches = sum([len(df.index) for df in grid])
+        num_patches = sum([len(df.index) for df in grid.values()])
```
mlubej, line 50: `len` works also on the df itself, so it could be just 

```suggestion
        num_patches = sum(map(len, grid.values()))
```

mlubej, line 50: `len` works also on the df itself, so it could be just 

```suggestion
        num_patches = sum(map(len, grid.values()))
```

eogrow/cli.py
```diff
@@ -85,7 +85,7 @@ def main(
         for raw_config in raw_configs:
             config = interpret_config_from_dict(raw_config, cli_variable_mapping)
             if test_patches:
-                config["patch_list"] = list(test_patches)
+                config["test_on"] = list(test_patches)
```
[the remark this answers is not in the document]
zigaLuksic: I want to discourage the use of this parameter outside of testing. I was thinking of even hiding this parameter from the users and have it only set via `-t` flag, but not sure i wanna do more magic than necessary. Filtration should be done through patch list files, not indices inside the config.

If you prefer the `subset` I guess we could use `test_subset`.

eogrow/pipelines/merge_samples.py
zigaLuksic: yes, but to be fair it wasn't trivial before either :)

mlubej: Before, I had to look at the index I was using, and find the eopatch name corresponding to that index in the grid file. How would one get the eopatch now? e.g. if I have an eopatch name and I want the index for that name, but there is some filtration going on.

zigaLuksic: it should correspond to the index of the name in the `finished.json` I believe

eogrow/cli.py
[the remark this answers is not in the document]
zigaLuksic: done

mlubej (approved):

Issue #160, opened 2022-12-29
[FEAT] Default `tiff_folder_key` parameter for ImportTiffPipeline


mlubej:
Usually we treat the `input-data` folder as the default when running some pipelines, where this is the location of input data to be used in the specific pipeline. 

Usually it was not the case that one had to provide specific folder keys for `input-data`, so perhaps I would suggest to use the `input-data` location as a default in cases where the `tiff_folder_key` parameter is not provided when running the  `ImportTiffPipeline`. This would make the `tiff_folder_key` parameter optional.

IMHO since the input-data is always there and is just one, there is no need to explicitly provide it as a folder key.

zigaLuksic:
Agreed. Prepare a MR or remind me if i forget :)

PR #161, opened 2023-01-03
Update parameter in InputTiffPipeline

mlubej:
Changes:
- updates to the `tiff_folder_key` parameter, makes it optional, the default being the input-data folder

Closes https://github.com/sentinel-hub/eo-grow/issues/160

### DIFF
--- a/eogrow/pipelines/import_tiff.py
+++ b/eogrow/pipelines/import_tiff.py
@@ -41,7 +41,12 @@ class ResizeSchema(BaseSchema):
 class ImportTiffPipeline(Pipeline):
     class Schema(Pipeline.Schema):
         output_folder_key: str = Field(description="The storage manager key of the output folder.")
-        tiff_folder_key: str = Field(description="The storage manager key of the folder containing the tiff to import.")
+        tiff_folder_key: Optional[str] = Field(
+            description=(
+                "The storage manager key of the folder containing the tiff to import. Defaults to the input-data"
+                " directory."
+            )
+        )
         input_filename: str = Field(description="Name of tiff file to import.")
         output_feature: Feature = Field(description="Feature containing the imported tiff information.")
         no_data_value: float = Field(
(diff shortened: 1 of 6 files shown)


eogrow/pipelines/import_tiff.py
```diff
@@ -41,7 +41,12 @@ def parameters_parser(cls, v: Tuple[ResizeParam, Tuple[float, float]]) -> Tuple[
 class ImportTiffPipeline(Pipeline):
     class Schema(Pipeline.Schema):
         output_folder_key: str = Field(description="The storage manager key of the output folder.")
-        tiff_folder_key: str = Field(description="The storage manager key of the folder containing the tiff to import.")
+        tiff_folder_key: Optional[str] = Field(
```
zigaLuksic, line 44: Knowledge from the high wizards:

In the definition of the storage manager, these pre-defined folders are given as:
```python
PRESET_FOLDERS: Dict[str, str] = {"logs": "logs", "input_data": "input-data", "cache": "cache"}
```

You can set the default to `input-data` and get rid of almost all the other changes to the code :)

tests/test_config_files/import_tiff/import_tiff_resized_scale_factors.json
```diff
@@ -1,7 +1,6 @@
 {
   "pipeline": "eogrow.pipelines.import_tiff.ImportTiffPipeline",
   "**global_config": "${config_path}/../global_config.json",
-  "tiff_folder_key": "input_data",
```
zigaLuksic, line 4: don't remove it everywhere. We want tests to detect a validation problem if someone accidentally deletes the key (yeah yeah i know it would probably crash elsewhere).

eogrow/pipelines/import_tiff.py
mlubej: done.

tests/test_config_files/import_tiff/import_tiff_resized_scale_factors.json
mlubej: done. removed it in half the cases.

PR #164, opened 2023-01-09
Restore tests for grid splitting pipeline

zigaLuksic:
They got removed when removing the old grid-switching, entirely my fault.
The results are slightly adjusted because the rng seeds changed a bit.


tests/test_pipelines/test_split_grid.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.parametrize("experiment_name", ["split_batch", "split_utm"])
+def test_grid_splitting(experiment_name, folders, batch_grid, test_storage_manager):
+    run_and_test_pipeline(experiment_name, **folders, folder_key="output_folder")
+
+    # check the output file is compatible with custom grid area managers
+    new_area_manager = CustomGridAreaManager.from_raw_config(
+        {"grid_folder_key": "temp", "grid_filename": "new_grid.gpkg", "name_column": "eopatch_name"},
+        storage=test_storage_manager,
+    )
+    new_area_manager.get_grid()
```
mlubej, line 40: Just to confirm, this just tests if grids, which were produced with the batch/utm process are successfully loaded with the CustomGridAM (meaning that they conform to all the requirements as well)?

mlubej (approved):

PR #167, opened 2023-01-09
Some minor polish

zigaLuksic:
Mostly related to docstrings.

I guess it's time to merge the big one afterwards.

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -28,7 +28,7 @@ class AreaSchema(BaseSchema):
     )
     simplification_factor: Optional[float] = Field(
         description=(
-            "Tolerance factor (in CRS units) for how much the buffered area geometry is simplified before splitting."
+            "Tolerance factor (in CRS units) for simplifying the buffered area geometry before splitting it."
         ),
     )
 
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -11,9 +11,9 @@ from pydantic import Field
 
 from sentinelhub import CRS, BBox, Geometry
 
+from ...types import PatchList
 from ...utils.eopatch_list import load_eopatch_names
 from ...utils.fs import LocalFile
-from ...utils.types import PatchList
 from ..base import EOGrowObject
 from ..schemas import BaseSchema, ManagerSchema
 from ..storage import StorageManager
@@ -24,14 +24,11 @@ LOGGER = logging.getLogger(__name__)
 class AreaSchema(BaseSchema):
     filename: str
     buffer: Optional[float] = Field(
-        description=(
-            "Buffer that will be applied to AOI geometry. Buffer has to be in the same units as AOI CRS. "
-            "In case buffer is too small, relatively to AOI size, it won't have any effect."
-        ),
+        description="Buffer that will be applied to AOI geometry. Buffer has to be in the same units as AOI CRS.",
     )
     simplification_factor: Optional[float] = Field(
         description=(
-            "A tolerance factor in CRS units how much the buffered area geometry will be simplified before splitting."
+            "Tolerance factor (in CRS units) for how much the buffered area geometry is simplified before splitting."
         ),
     )
 
@@ -55,7 +52,7 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
 
     def __init__(self, config: Schema, storage: StorageManager):
         """
-        :param config: A configuration file
+        :param config: The configuration schema
         :param storage: An instance of StorageManager class
         """
         super().__init__(config)
@@ -92,7 +89,7 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
         """
 
     def _load_grid(self, grid_path: str) -> Dict[CRS, gpd.GeoDataFrame]:
-        """A method that loads bounding box grid saved in a cache folder"""
+        """A method that loads the bounding box grid saved in the cache folder."""
         LOGGER.info("Loading grid from %s", grid_path)
 
         grid = {}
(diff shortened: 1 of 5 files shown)


eogrow/core/area/base.py
```diff
@@ -24,14 +24,11 @@
...
-        description=(
-            "Buffer that will be applied to AOI geometry. Buffer has to be in the same units as AOI CRS. "
-            "In case buffer is too small, relatively to AOI size, it won't have any effect."
-        ),
+        description="Buffer that will be applied to AOI geometry. Buffer has to be in the same units as AOI CRS.",
     )
     simplification_factor: Optional[float] = Field(
         description=(
-            "A tolerance factor in CRS units how much the buffered area geometry will be simplified before splitting."
+            "Tolerance factor (in CRS units) for how much the buffered area geometry is simplified before splitting."
```
mlubej, line 31:
```suggestion
            "Tolerance factor (in CRS units) for simplifying the buffered area geometry before splitting it."
```

eogrow/core/area/base.py
```diff
@@ -92,7 +89,7 @@ def _create_grid(self) -> Dict[CRS, gpd.GeoDataFrame]:
         """
 
     def _load_grid(self, grid_path: str) -> Dict[CRS, gpd.GeoDataFrame]:
-        """A method that loads bounding box grid saved in a cache folder"""
+        """A method that loads the bounding box grid saved in the cache folder."""
```
mlubej, line 92:
```suggestion
        """A method that loads the bounding box grid from the cache folder."""
```

eogrow/core/area/batch.py
```diff
@@ -30,15 +30,16 @@ class Schema(BaseAreaManager.Schema):
         )
         resolution: float = Field(
```
mlubej, line 31: why not rename this parameter to `tiling_grid_id`, since the user needs to provide the ID anyway?

eogrow/core/area/batch.py
```diff
@@ -30,15 +30,16 @@ class Schema(BaseAreaManager.Schema):
...
+                " chosen tiling_grid_id."
             )
         )
         tile_buffer_x: int = Field(0, description="Number of pixels for which to buffer each tile left and right.")
         tile_buffer_y: int = Field(0, description="Number of pixels for which to buffer each tile up and down.")
         batch_id: Optional[str] = Field(
             description=(
-                "An ID of a batch job for this pipeline. If it is given the pipeline will just monitor the "
-                "existing batch job. If it is not given it will create a new batch job."
+                "An ID of a batch job that defines the AOI. Not required when using BatchDownloadPipeline,"
```
mlubej, line 41:
```suggestion
                "ID of the batch job that defines the AOI. Not required when using BatchDownloadPipeline,"
```

eogrow/core/area/batch.py
```diff
@@ -91,8 +92,8 @@ def _create_grid(self) -> Dict[CRS, GeoDataFrame]:
         return grid
 
     def _verify_batch_request(self, batch_request: BatchRequest) -> None:
-        """Verifies that given batch request has finished and that it has the same tiling grid parameters as
-        they are written in the config.
+        """Verifies that the given batch request has finished and that it has the same tiling grid parameters as
```
mlubej, line 95: perhaps this makes the reference to the request object more prominent 

```suggestion
        """Verifies that the given batch request has finished and that it contains the same tiling grid parameters as
```

eogrow/core/pipeline.py
```diff
@@ -104,18 +105,19 @@ def get_patch_list(self) -> PatchList:
...
             LOGGER.info("%s, %d / %d remaining", skip_message, len(filtered_patch_list), len(patch_list))
 
-            patch_list = filtered_patch_list
+            return filtered_patch_list
 
         return patch_list
 
     def filter_patch_list(self, patch_list: PatchList) -> PatchList:
-        """Overwrite this method to specify which EOPatches should be filtered with `skip_existing`"""
+        """Specifies which EOPatches should be filtered when `skip_existing` is enabled."""
```
mlubej, line 113:
```suggestion
        """Specifies which EOPatches should be skipped when `skip_existing` is enabled."""
```

eogrow/core/pipeline.py
```diff
@@ -200,7 +202,7 @@ def run_execution(
         return successful, failed, execution_results
 
     def run(self) -> None:
-        """Call this method to run any pipeline"""
+        """The main method of pipeline execution. Sets up logging and runs the pipeline procedure."""
```
mlubej, line 205:
```suggestion
        """The main method for pipeline execution. It sets up logging and runs the pipeline procedure."""
```

eogrow/core/pipeline.py
```diff
@@ -248,11 +250,11 @@ def run_procedure(self) -> Tuple[List[str], List[str]]:
 
         By default, builds the workflow by using a `build_workflow` method, which must be additionally implemented.
 
-        :return: A list of successfully executed EOPatch names and a list of unsuccessfully executed EOPatch names
+        :return: A list of successful executions and a list of unsuccessful executions
```
mlubej, line 253:
```suggestion
        :return: A pair of lists representing successful and unsuccessful executions.
```

eogrow/core/pipeline.py
```diff
@@ -248,11 +250,11 @@ def run_procedure(self) -> Tuple[List[str], List[str]]:
...
         """
         if not hasattr(self, "build_workflow"):
             raise NotImplementedError(
-                "Default implementation of run_procedure method requires implementation of build_workflow method"
+                "Default implementation of run_procedure method requires implementation of build_workflow method."
```
mlubej, line 257:
```suggestion
                "Default implementation of the `run_procedure` method requires implementation of the `build_workflow` method."
```

eogrow/core/area/batch.py
zigaLuksic: oh no, i borked that up. It is not the ID of the resolution but the actual resolution, it's just constrained :see_no_evil: good catch, will fix

mlubej: Makes sense, but would it still be better if one would provide the grid ID instead of the resolution? seems more intuitive and direct, or does it complicate other things down the line?

zigaLuksic: check L28 ;)

mlubej: 🙈

mlubej (approved):

PR #168, opened 2023-01-10
Area manager rework

zigaLuksic:
Merges the EOPatch and Area managers into a single AreaManager class


mlubej:
Anything to do here or just a formality?

zigaLuksic:
> Anything to do here or just a formality?

more or less a formality :) you are free to double-check everything if you like

mlubej (approved):

PR #174, opened 2023-01-23
PR for the bugfix : process killed while executing ClassificationPredictionPipeline

aashishd:
process killed while running eogrow.pipelines.prediction.ClassificationPredictionPipeline

To Reproduce
The issue is reproducible when the property "label_encoder_filename" is specified in the cofiguration.
In such case, the LabelEncoder (LE) which was used in training step is used to decode the labels.

This happens in the eogrow.pipelines.prediction.ClassificationPredictionTask line 163 where we try to transform the form of the feature which accepts predictions to be a column feature. But because we are using a LE for decoding, it converts the column predictions array to row array. This causes the function transform_to_feature_form to just fail without any issue.

Run the eogrow.pipelines.training.ClassificationTrainingPipeline with a preprocessing.label_encoder_filename specified
Once the model is trained, run eogrow.pipelines.prediction.ClassificationPredictionPipeline and use the same label_encoder_filename which was used in training to decode
Expected behavior

Environment

Python 3.9

Stack trace or screenshots

The issue doesn't give any stack trace, the process is killed with just a small warning as follows :
............

2023-01-19 15:34:11,381 graphviz._tools DEBUG    deprecate positional args: graphviz.graphs.BaseGraph.__init__(['comment', 'filename', 'directory', 'format', 'engine', 'encoding', 'graph_attr', 'node_attr', 'edge_attr', 'body', 'strict'])
2023-01-19 15:34:11,382 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.from_file(['directory', 'format', 'engine', 'encoding', 'renderer', 'formatter'])
2023-01-19 15:34:11,383 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.__init__(['filename', 'directory', 'format', 'engine', 'encoding'])
2023-01-19 15:34:11,383 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.save(['directory'])
2023-01-19 15:35:16,970 py.warnings  WARNING  /Users/dhiman/envs/uc_land-cover/lib/python3.9/site-packages/sklearn/preprocessing/_label.py:155: DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples, ), for example using ravel().
  y = column_or_1d(y, warn=True)
Desktop (please complete the following information):

OS: Docker container
Version
sentinelhub>=3.6.3
eo-learn==1.3.1
eo-grow==1.3.3
scikit-learn==1.2.0
Solution
I was able to fix the issue by moving the line where you add new axis (i.e. line 163 in file eogrow.pipelines.prediction => predictions = predictions[..., np.newaxis]) after the labels are decoded. I tried to push my branch with the fix to create a PR but I don't have the rights to the repo.


zigaLuksic:
I moved the merge branch from `main` to `develop` and ran tests.

aashishd:
Is there am ETA on when a new version of eo-grow would be released which includes this change?

zigaLuksic:
> Is there am ETA on when a new version of eo-grow would be released which includes this change?

you're in luck, this week :)
However, the release will be fairly codebreaking. The area manager has been changed and this caused some larger refactoring in the rest of the code. I can send you a small guide for updating via e-mail.

PR #177, opened 2023-01-23
Move patch_list filtration to grid

zigaLuksic:
Still wont support -t filtration, but it does filter the grid.

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -76,6 +76,23 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
 
         grid = self._create_grid()
         self._save_grid(grid, grid_path)
+
+        if self.config.patch_list is not None:
+            folder_path = self.storage.get_folder(self.config.patch_list.input_folder_key)
+            patch_list_path = fs.path.join(folder_path, self.config.patch_list.filename)
+            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_list_path))
+
+            for crs, geoms in grid.items():
+                grid[crs] = geoms[geoms[self.NAME_COLUMN].isin(relevant_patches)]
+
+            grid = {crs: geoms for crs, geoms in grid.items() if not geoms.empty}
+
+            num_geoms = sum(map(len, grid.values()))
+            if len(relevant_patches) != num_geoms:
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


eogrow/core/area/base.py
```diff
@@ -121,20 +138,12 @@ def get_grid_cache_filename(self) -> str:
...
     def get_patch_list(self) -> PatchList:
         """Returns a list of eopatch names and appropriate BBoxes."""
-        relevant_patches = None
-        if self.config.patch_list is not None:
-            folder_path = self.storage.get_folder(self.config.patch_list.input_folder_key)
-            patch_list_path = fs.path.join(folder_path, self.config.patch_list.filename)
-            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_list_path))
 
         named_bboxes = []
         for crs, grid in self.get_grid().items():
```
mlubej, line 143: I remember us saying something about the possibility of having a flag in the `get_grid()` function, to still enable the user/developer getting the original grid/patches, regardless of the patch list being present or not.

something along the lines of:

```python
def get_grid(self, filtered: bool = True) -> Dict[CRS, gpd.GeoDataFrame]:
```

or the default could even be false, to retain the same behavior as in the current branch

zigaLuksic: i actually had that in mind, all the way up to the point where i got to doing it, and then promptly forgot.

added now

eogrow/core/area/base.py
```diff
@@ -121,20 +138,12 @@ def get_grid_cache_filename(self) -> str:
 
     def get_patch_list(self) -> PatchList:
```
mlubej, line 139: Would this function also benefit from having the flag?

I guess it's better to keep it simple and have a 1 to 1 relation to what is used. All patches can be obtained manually in case they are needed.

zigaLuksic: I think even the grid doesnt necessarily need to have the flag. Sometimes it's better to not give the user too many options to think about. The patch list should be filtered, and the user should think real hard on why someone would specify filtering, but then have it ignored.

If filtering is not allowed for a pipeline, one can set the type in the schema to `Literal[None]` similar to what is sometimes done for `skip_existing`.

It is also easier to add functionality that to take it away, so I'd prefer to add it later if need shows.

mlubej (approved):

PR #180, opened 2023-01-24
Add tests for tasks spatial

jgersak:
add test for get_array_slices

### DIFF
--- a/tests/test_tasks/test_spatial.py
+++ b/tests/test_tasks/test_spatial.py
@@ -79,8 +79,8 @@ pytestmark = pytest.mark.fast
             (slice(10, 190, None), slice(20, 90, None)),
             id="test slice_bbox included",
         ),
-        # This test is specific to the current implementation, and serves only to detect unplanned changes.
         pytest.param(
+                # Test is specific to current implementation, serves only to detect unplanned changes.
             {
                 "bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
                 "slice_bbox": BBox(((731000, 4400500), (733000, 4401500)), crs=CRS("32638")),
--- a/tests/test_tasks/test_spatial.py
+++ b/tests/test_tasks/test_spatial.py
@@ -1,0 +1,93 @@
+from typing import Any, Dict, Tuple
+
+import pytest
+
+from sentinelhub import CRS, BBox
+
+from eogrow.tasks.spatial import get_array_slices
+
+
+@pytest.mark.parametrize(
+    "input_params, expected",
+    [
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "slice_bbox": BBox(((729480.0, 4390255.0), (730480.0, 4391255.0)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(0, 100, None), slice(0, 100, None)),
+        ),
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "slice_bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(0, 264, None), slice(0, 121, None)),
+        ),
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "slice_bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "resolution": (10, 10),
+                "limit_x": (50, 100),
+                "limit_y": (100, 200),
+            },
+            (slice(50, 100, None), slice(100, 121, None)),
+        ),
+        (
+            {
+                "bbox": BBox(bbox=[500, 500, 600, 600], crs=CRS.WGS84),
+                "slice_bbox": BBox(bbox=[500, 500, 600, 600], crs=CRS.WGS84),
+                "resolution": (10, 10),
+            },
+            (slice(0, 10, None), slice(0, 10, None)),
+        ),
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "slice_bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "size": (10, 10),
+            },
+            (slice(0, 10, None), slice(0, 10, None)),
+        ),
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "slice_bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+                "size": (10, 10),
+                "limit_x": (50, 100),
+                "limit_y": (100, 200),
+            },
+            (slice(50, 10, None), slice(100, 10, None)),
+        ),
+    ],
+)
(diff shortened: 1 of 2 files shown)


tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,93 @@
...
+
+from eogrow.tasks.spatial import get_array_slices
+
+
+@pytest.mark.parametrize(
+    "input_params, expected",
+    [
+        (
+            {
+                "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
```
zigaLuksic, line 15: Its hard to figure out wether the expected result really is what is expected, because BBoxes have such convoluted coordinates. Would be easier if they were simplified a bit, for instance make all points a bit above `(730000, 4400000)` so that we can focus on the changes in the lower digits without a calculator.

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,93 @@
...
+            "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+            "slice_bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+            "resolution": (10, 10),
+            "size": (10, 10),
+        },
+        {
+            "bbox": BBox(((729480.0, 4390045.0), (732120.0, 4391255.0)), crs=CRS("32638")),
+            "slice_bbox": BBox(((729480.5, 4390045.5), (732120.0, 4391255.0)), crs=CRS("32638")),
+            "resolution": (10, 10),
+        },
```
zigaLuksic, line 88: add comments (or even better, a test case name) to these, so that it is clear WHY they should fail

jgersak: Add test ids which add reason why test should fail, add it to working cases as well.

zigaLuksic: Currently it is difficult to see which IDs belong to which cases. Consider using `pytest.param` (see [here](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids)) or an id-parser like we did in tests for `eolearn.core.utils.parsing`.

tests/test_tasks/test_spatial.py
zigaLuksic: good, now make some test cases less regular (make them wide 1110 meters instead of 1000 etc.)

tests/test_tasks/test_spatial.py
jgersak: done

tests/test_tasks/test_spatial.py
jgersak: added

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+    "input_params, expected",
+    [
+        pytest.param(
+            {
+                "bbox": BBox(((730000, 4400000), (731000, 4401000)), crs=CRS("32638")),
+                "slice_bbox": BBox(((730000, 4400000), (731000, 4401000)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(0, 100, None), slice(0, 100, None)),
+            id="clean test",
```
zigaLuksic, line 20:
```suggestion
            id="fully covered",
```

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+        ),
+        pytest.param(
+            {
+                "bbox": BBox(((730000, 4400000), (731000, 4401000)), crs=CRS("32638")),
+                "slice_bbox": BBox(((730000, 4400000), (730990, 4400650)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(0, 99, None), slice(0, 65, None)),
+            id="less regular test",
```
zigaLuksic, line 29:
```suggestion
            id="contained",
```

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+        ),
+        pytest.param(
+            {
+                "bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
+                "slice_bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(0, 200, None), slice(0, 100, None)),
+            id="random test",
```
zigaLuksic, line 38: doesnt this one match the first test? except it's not symmetrical :thinking: 
I would keep this one and throw the first one out since they both test the `fully covered` test case

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+                "slice_bbox": BBox(bbox=[500, 500, 600, 600], crs=CRS.WGS84),
+                "resolution": (10, 10),
+            },
+            (slice(0, 10, None), slice(0, 10, None)),
+            id="test WGS84",
+        ),
+        pytest.param(
+            {
+                "bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
+                "slice_bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
```
zigaLuksic, line 63: there is no reason to again have a fully covered test. Try adding a test case where the bboxes don't align in the lower left corner.

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+                "size": (10, 10),
+                "limit_x": (50, 100),
+                "limit_y": (100, 200),
+            },
+            (slice(50, 10, None), slice(100, 10, None)),
```
zigaLuksic, line 77: the answer here is nonsensical, a `50:10` slice is nonsense.
the `size` is clearly too small (expects 10x10 pixel data) for the `limit_x` and `limit_y`. Consider if an error should be raised in this case.

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,117 @@
...
+def test_get_array_slices(input_params: Dict[str, Any], expected: Tuple[slice, slice]) -> None:
+    assert get_array_slices(**input_params) == expected
+
+
+@pytest.mark.parametrize(
+    "input_params",
+    [
+        pytest.param(
+            {
+                "bbox": BBox(((730000, 4400000), (732000, 4401000)), crs=CRS("32638")),
```
zigaLuksic, line 91: this BBox repeats itself often, consider extracting it as a constant (at least for the `test_get_array_slices_invalid_input`)

tests/test_tasks/test_spatial.py
jgersak: updated with suggestions from mattermost

tests/test_tasks/test_spatial.py
jgersak: done

tests/test_tasks/test_spatial.py
jgersak: updated

tests/test_tasks/test_spatial.py
jgersak: done

tests/test_tasks/test_spatial.py
```diff
@@ -0,0 +1,144 @@
...
+                "slice_bbox": BBox(((730100, 4400100), (731900, 4400800)), crs=CRS("32638")),
+                "resolution": (10, 10),
+            },
+            (slice(10, 190, None), slice(20, 90, None)),
+            id="test slice_bbox included",
+        ),
+        # This test is specific to the current implementation, and serves only to detect unplanned changes.
+        pytest.param(
```
zigaLuksic, line 83:
```suggestion
        pytest.param(
                # Test is specific to current implementation, serves only to detect unplanned changes.
```
Moved so that it is more clear which test it refers to
i shortened the message so that black wont complain if you merge from github

zigaLuksic (approved):

PR #182, opened 2023-01-24
Remove MappingPipeline

jgersak:
remove MappingPipeline
remove MappingTask
update accordingly


tests/test_core/test_schemas.py
```diff
@@ -4,13 +4,12 @@
 from eogrow.core.schemas import build_schema_template
 from eogrow.core.storage import StorageManager
 from eogrow.pipelines.export_maps import ExportMapsPipeline
-from eogrow.pipelines.mapping import MappingPipeline
```
zigaLuksic, line 7: switch this for `eogrow.pipelines.zipmap import ZipMapPipeline` and add it to the test. I want at least two pipelines in there :)

jgersak: done

PR #183, opened 2023-01-25
create a naming distinction between patch_list and patch_names

zigaLuksic:
There wasn't much to change, the code abstraction is good :D

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -80,17 +80,17 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
         if filtered and self.config.patch_names is not None:
             folder_path = self.storage.get_folder(self.config.patch_names.input_folder_key)
             patch_names_path = fs.path.join(folder_path, self.config.patch_names.filename)
-            relevant_patches = set(load_names(self.storage.filesystem, patch_names_path))
+            relevant_patch_names = set(load_names(self.storage.filesystem, patch_names_path))
 
             for crs, geoms in grid.items():
-                grid[crs] = geoms[geoms[self.NAME_COLUMN].isin(relevant_patches)]
+                grid[crs] = geoms[geoms[self.NAME_COLUMN].isin(relevant_patch_names)]
 
             grid = {crs: geoms for crs, geoms in grid.items() if not geoms.empty}
 
             num_geoms = sum(map(len, grid.values()))
-            if len(relevant_patches) != num_geoms:
+            if len(relevant_patch_names) != num_geoms:
                 raise ValueError(
-                    f"Filtration done with {len(relevant_patches)} unique names, but {num_geoms} patches were found"
+                    f"Filtration done with {len(relevant_patch_names)} unique names, but {num_geoms} patches were found"
                 )
 
         return grid
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -12,7 +12,7 @@ from pydantic import Field
 from sentinelhub import CRS, BBox, Geometry
 
 from ...types import PatchList
-from ...utils.eopatch_list import load_eopatch_names
+from ...utils.eopatch_list import load_names
 from ...utils.fs import LocalFile
 from ..base import EOGrowObject
 from ..schemas import BaseSchema, ManagerSchema
@@ -80,7 +80,7 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
         if filtered and self.config.patch_names is not None:
             folder_path = self.storage.get_folder(self.config.patch_names.input_folder_key)
             patch_names_path = fs.path.join(folder_path, self.config.patch_names.filename)
-            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_names_path))
+            relevant_patches = set(load_names(self.storage.filesystem, patch_names_path))
 
             for crs, geoms in grid.items():
                 grid[crs] = geoms[geoms[self.NAME_COLUMN].isin(relevant_patches)]
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -42,7 +42,7 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
     NAME_COLUMN = "eopatch_name"
 
     class Schema(ManagerSchema):
-        relevant_patch_names: Optional[PatchListSchema] = Field(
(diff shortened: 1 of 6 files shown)


eogrow/core/area/base.py
```diff
@@ -42,8 +42,8 @@ class BaseAreaManager(EOGrowObject, metaclass=ABCMeta):
     NAME_COLUMN = "eopatch_name"
 
     class Schema(ManagerSchema):
-        patch_list: Optional[PatchListSchema] = Field(
-            description="Names of EOPatches to keep when filtering in `get_names_and_bboxes` method."
+        patch_names_list: Optional[PatchListSchema] = Field(
```
mlubej, line 45: I find "list" here redundant, could just be "patch_names".

However, it's OK to have `patch_list` then for the `List[Tuple[str, BBox]]`.

eogrow/core/area/base.py
```diff
@@ -77,10 +77,10 @@ def get_grid(self, filtered: bool = True) -> Dict[CRS, gpd.GeoDataFrame]:
...
         self._save_grid(grid, grid_path)
 
-        if filtered and self.config.patch_list is not None:
-            folder_path = self.storage.get_folder(self.config.patch_list.input_folder_key)
-            patch_list_path = fs.path.join(folder_path, self.config.patch_list.filename)
-            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_list_path))
+        if filtered and self.config.patch_names_list is not None:
+            folder_path = self.storage.get_folder(self.config.patch_names_list.input_folder_key)
+            patch_names_path = fs.path.join(folder_path, self.config.patch_names_list.filename)
+            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_names_path))
```
mlubej, line 83: `relevant_patches` should be more in line with `patch_names`/`patch_names_list`, because it's also just a list of names, right?

eogrow/utils/eopatch_list.py
```diff
@@ -11,16 +11,16 @@
 from ..types import PatchList
 
 
-def save_eopatch_names(filesystem: FS, file_path: str, eopatch_list: List[str]) -> None:
+def save_eopatch_names(filesystem: FS, file_path: str, eopatch_names: List[str]) -> None:
```
mlubej, line 14: should we use consistency also here and use `patch_names`/`patch_names_list`?

eogrow/core/area/base.py
zigaLuksic: i was afraid that `patch_names` carries a connotation that it is a list of 'how to name patches'

to be fair `patch_names_list` is not much better in that regard, we're just used to the `patch_list` name.

Also related to your comment bellow, how do you feel about having this parameter named `relevant_patch_names`?

mlubej: From this point of view, assuming that `patch` means a named bbox `Tuple[str, BBox]`, then it's fine if we use:
- `patch_names` and `relevant_patch_names`/`filtered_patch_names`/`patch_names_subset` for `List[str]`
- `patch_list` for `List[Tuple[str, BBox]]`

I feel the `names` part already suggests a list and is sufficient.

> Also related to your comment bellow, how do you feel about having this parameter named relevant_patch_names?

I'm assuming this is related to [eogrow/core/area/base.py](https://github.com/sentinel-hub/eo-grow/pull/183/files/3330fcd411377dfde1fbafb2ec4f8b0c084d7bf4#diff-a664a1d150adb15d5edbfaeda2b2ffbef2e905d1b9ce5449670f5f88406b1056) L83?

zigaLuksic: I am leaning towards `patch_names_subset` since we have a `test_subset`. Although `relevant_patch_names` sounds nicer and is very clear

eogrow/utils/eopatch_list.py
zigaLuksic: not sure, technically this is more general. If anything, it should be changed to 'save_execution_names` and `load_eopatch_names` but that is just confusing :see_no_evil: 

Perhaps just `save_names` and `load_names`?

eogrow/core/area/base.py
zigaLuksic: I went with `relevant_patch_names`, how does the code look to ya?

eogrow/utils/eopatch_list.py
mlubej: Sorry, I was talking about the argument `eopatch_names` here.

But your comment is relevant, so if it should be more general, then I like your last suggestions.

eogrow/core/area/base.py
mlubej: Wait, I must have misunderstood. I didn't realize you meant to use `relevant_patch_names` _instead_ of `patch_names`. My comments for `relevant_*` were for L83. I think that in this case `patch_names` was completely satisfactory, especially together with the description.

zigaLuksic: changed to `patch_names`

eogrow/utils/eopatch_list.py
zigaLuksic: changed to `save_names` and `load_names`

eogrow/core/area/base.py
```diff
@@ -77,10 +77,10 @@ def get_grid(self, filtered: bool = True) -> Dict[CRS, gpd.GeoDataFrame]:
...
         self._save_grid(grid, grid_path)
 
-        if filtered and self.config.patch_list is not None:
-            folder_path = self.storage.get_folder(self.config.patch_list.input_folder_key)
-            patch_list_path = fs.path.join(folder_path, self.config.patch_list.filename)
-            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_list_path))
+        if filtered and self.config.patch_names is not None:
+            folder_path = self.storage.get_folder(self.config.patch_names.input_folder_key)
+            patch_names_path = fs.path.join(folder_path, self.config.patch_names.filename)
+            relevant_patches = set(load_eopatch_names(self.storage.filesystem, patch_names_path))
```
mlubej: Ok, this is now the only thing left, since it would make me think that `relevant_patches` is of the `List[Tuple[str, BBox]]` type, but is in fact of the `List[str]` type (`Set[str]` really, but still).  So here I would suggest to use any of the following (in that order):

1. `patch_names_subset`
2. `filtered_patch_names`
3. `relevant_patch_names`

mlubej (approved):

PR #193, opened 2023-02-20
Feat/temporal rasterisation

batic:
Load (and save) eopatch timestamps if rasterisation for temporal features is requested.

### DIFF
--- a/eogrow/pipelines/rasterize.py
+++ b/eogrow/pipelines/rasterize.py
@@ -169,10 +169,13 @@ class RasterizePipeline(Pipeline):
             )
             data_preparation_node = EONode(import_task, inputs=[create_node])
         else:
+            features = [self.vector_feature, FeatureType.BBOX]
+            if self.vector_feature[0].is_temporal():
+                features.append(FeatureType.TIMESTAMP)
             input_task = LoadTask(
                 self.storage.get_folder(self.config.input_folder_key),
                 filesystem=self.storage.filesystem,
-                features=[self.vector_feature, FeatureType.BBOX],
+                features=features,
             )
             data_preparation_node = EONode(input_task)
 
--- a/eogrow/pipelines/rasterize.py
+++ b/eogrow/pipelines/rasterize.py
@@ -106,12 +106,17 @@ class RasterizePipeline(Pipeline):
         if len({self._is_temporal(c.output_feature) for c in self.config.columns}) != 1:
             raise ValueError("All output features have to be either temporal or timeless!")
 
+        output_is_temporal = self._is_temporal(self.config.columns[0].output_feature)
         if isinstance(self.config.vector_input, str):
             self.filename = self._parse_input_file(self.config.vector_input)
-            output_is_temporal = self._is_temporal(self.config.columns[0].output_feature)
             feature_type = FeatureType.VECTOR if output_is_temporal else FeatureType.VECTOR_TIMELESS
             self.vector_feature = feature_type, f"TEMP_{uuid.uuid4().hex}"
         else:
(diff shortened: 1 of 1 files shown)


eogrow/pipelines/rasterize.py
```diff
@@ -252,5 +255,9 @@ def _get_dataset_path(self, filename: str) -> str:
     def _get_output_features(self) -> List[FeatureSpec]:
         """Lists all features that are to be saved upon the pipeline completion"""
         features: List[FeatureSpec] = [FeatureType.BBOX]
+
+        if self.vector_feature[0].is_temporal():
```
mlubej, line 259: As you mentioned, you're bothered by the `self.vector_feature[0].is_temporal()`, which I agree with. The only thing I can come up with is defining the following

```python
f_type, _ = self.vector_feature
```

and then using `f_type.is_temporal()`. No need to take this into account though. Approving, but I'm letting you decide.

mlubej (approved):

eogrow/pipelines/rasterize.py
zigaLuksic: another option is to have a static method `vector_feature_is_temporal(feature)` and hide that in there.
Sure it's a one-liner, but if it makes the code easier to read...

eogrow/pipelines/rasterize.py
```diff
@@ -252,5 +255,9 @@ def _get_dataset_path(self, filename: str) -> str:
...
+            features.extend([FeatureType.TIMESTAMP])
```
zigaLuksic, line 260: any reason for extend instead of `append`?

batic: none whatsoever

eogrow/pipelines/rasterize.py
batic: can do oneliner static method, but wasn't sure where to add it; should it be in the class, or in utils, or ....

zigaLuksic: I would either add it as part of the class, or add it as a utility function at the end of the file with a `_` prefix so it's not in the docs.

eogrow/pipelines/rasterize.py
```diff
@@ -103,12 +103,12 @@ def __init__(self, *args: Any, **kwargs: Any):
...
         self.filename: Optional[str] = None
 
-        if len({c.output_feature[0].is_temporal() for c in self.config.columns}) != 1:
+        if len({self._is_temporal(c.output_feature) for c in self.config.columns}) != 1:
             raise ValueError("All output features have to be either temporal or timeless!")
 
         if isinstance(self.config.vector_input, str):
             self.filename = self._parse_input_file(self.config.vector_input)
-            output_is_temporal = self.config.columns[0].output_feature[0].is_temporal()
+            output_is_temporal = self._is_temporal(self.config.columns[0].output_feature)
```
zigaLuksic, line 111: shouldn't this check ALL output features?

zigaLuksic (approved):

eogrow/pipelines/rasterize.py
batic: I agree.

It is partially already done; there is a check for input features to be all temporal or timeless. So the final (missing) check is that output feature also conforms to input feature timeliness.

batic: Added the (quick) test, we can address how to check this properly (at schema validation phase?) in another issue.

zigaLuksic:
@mlubej any other suggestions?

mlubej:
> @mlubej any other suggestions?

No, good to go AFAIC.

Issue #172, opened 2023-01-19
process killed while running eogrow.pipelines.prediction.ClassificationPredictionPipeline


ashishdhiman-tomtom:
**process killed while running eogrow.pipelines.prediction.ClassificationPredictionPipeline**

**To Reproduce**
The issue is reproducible when the property "label_encoder_filename" is specified in the cofiguration.
In such case, the LabelEncoder (LE) which was used in training step is used to decode the labels.

This happens in the eogrow.pipelines.prediction.ClassificationPredictionTask line 163 where we try to transform the form of the feature which accepts predictions to be a column feature. But because we are using a LE for decoding, it converts the column predictions array to row array. This causes the function transform_to_feature_form to just fail without any issue. 

1. Run the eogrow.pipelines.training.ClassificationTrainingPipeline with a preprocessing.label_encoder_filename specified
2. Once the model is trained, run eogrow.pipelines.prediction.ClassificationPredictionPipeline and use the same label_encoder_filename which was used in training to decode

**Expected behavior**

**Environment**

Python 3.9

**Stack trace or screenshots**

The issue doesn't give any stack trace, the process is killed with just a small warning as follows : 
............
```
2023-01-19 15:34:11,381 graphviz._tools DEBUG    deprecate positional args: graphviz.graphs.BaseGraph.__init__(['comment', 'filename', 'directory', 'format', 'engine', 'encoding', 'graph_attr', 'node_attr', 'edge_attr', 'body', 'strict'])
2023-01-19 15:34:11,382 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.from_file(['directory', 'format', 'engine', 'encoding', 'renderer', 'formatter'])
2023-01-19 15:34:11,383 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.__init__(['filename', 'directory', 'format', 'engine', 'encoding'])
2023-01-19 15:34:11,383 graphviz._tools DEBUG    deprecate positional args: graphviz.sources.Source.save(['directory'])
2023-01-19 15:35:16,970 py.warnings  WARNING  /Users/dhiman/envs/uc_land-cover/lib/python3.9/site-packages/sklearn/preprocessing/_label.py:155: DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples, ), for example using ravel().
  y = column_or_1d(y, warn=True)
```

**Desktop (please complete the following information):**

 - OS: Docker container
[...]

PR #194, opened 2023-03-03
Bugfix export maps many tiff

aashishd:
PR contains two bug fixes. We have GEM issues for both of them.
**1. Bug fix in the prediction step**
   While using the LabelEncoder the the new axis added is removed so, the solution is to just add the new axis after wards
**2. Bug fix for multiple tiff files processing**
   There is a limit to the size of the command that can be executed on a unix based system. In case of too many files, this limit was exceeding. Then there's also a limit on how many files can stay open at once with a process in Unix bases system. So, the solution is to add two checks and create an intermediary vrt file, and then convert this vrt file to tiff using gdalwarp. I have tested it for an AOI with around 4000 files in Myanmar, and it works. Please test it on your side too.


aashishd:
The first bug fix mentioned is the solution to this issue : https://github.com/sentinel-hub/eo-grow/issues/172

eogrow/utils/map.py
```diff
@@ -163,7 +168,22 @@ def merge_tiffs(
     if dtype is not None:
         gdalwarp_options += f" -ot {GDAL_DTYPE_SETTINGS[dtype]}"
 
-    command = f"gdalwarp {gdalwarp_options} {' '.join(input_filenames)} {merged_filename}"
+    input_filelist = list(input_filenames)
+    command = f"gdalwarp {gdalwarp_options} {' '.join(input_filelist)} {merged_filename}"
+
+    if len(command) > SH_COMMAND_LIMIT or len(input_filelist) > OPEN_FILES_LIMIT:
+        # generate a text file with
```
batic, line 175: It might be a good idea to create the vrt file as temporary file that is cleaned up afterwards. Is there any reason why you'd like to have the file persisted?

eogrow/utils/map.py
```diff
@@ -163,7 +168,22 @@ def merge_tiffs(
...
+        merged_path = Path(merged_filename)
+        tile_list_path = str(merged_path.with_name(f"{merged_path.stem}_files.txt"))
```
batic, line 177: No need to cast paths to string; `open` works directly with paths, and when constructing `generate_vrt_command` and `command`, the path will be cast to string automatically.

eogrow/utils/map.py
```diff
@@ -163,7 +168,22 @@ def merge_tiffs(
...
+        vrt_file_path = str(merged_path.with_name(f"{merged_path.stem}_temp.vrt"))
+        with open(tile_list_path, "w") as file:
+            for string in input_filelist:
```
batic, line 180: Probably a personal preference, but this could be rewritten as 
```
with open(vrt_file_path, "w") as vrt_file:
    vrt_file.writelines([f"{file}\n" for file in input_filelist])
```

eogrow/utils/map.py
```diff
@@ -163,7 +168,22 @@ def merge_tiffs(
...
+                file.write(string + "\n")
+        generate_vrt_command = f"gdalbuildvrt {vrt_file_path} -input_file_list {tile_list_path}"
+
+        LOGGER.info(f"Command too big or Too many files to process. Creating an intermediary vrt: {vrt_file_path}")
```
batic, line 184: typo (`Too` -> `too`)

Probably more accurate description would be "Command or argument list too long. Creating an intermediary vrt: {vrt_file_path}."

eogrow/utils/map.py
```diff
@@ -7,6 +7,11 @@
 import subprocess
 from tempfile import NamedTemporaryFile
 from typing import Iterable, Literal, Optional
+from pathlib import Path
```
batic, line 10: `isort` complains about the order. Please run `pre-commit run --all-files` to fix.

eogrow/utils/map.py
aashishd: Generally when running using ray in a container, this file is cleaned when the container is killed but I've added code to do that.

eogrow/utils/map.py
aashishd: resolved

eogrow/utils/map.py
aashishd: resolved

eogrow/utils/map.py
aashishd: resolved

aashishd: The pre-commit run --all-files command ran successfully on my machine. Rebased the branch and pushed again. Please let me know if this looks okay now.

zigaLuksic:
Hi @aashishd, it looks like something went terribly wrong with your merge and i can't run our tests.

aashishd:
@zigaLuksic , I think there were some pending merge conflicts which I resolved, maybe try now ?

eogrow/utils/map.py
aashishd: resolved

eogrow/utils/map.py
batic: I actually like that vrt is persisted, and only the temp file that serves as input to build vrt is temporary. This way one can still inspect the vrt.

PR #196, opened 2023-03-03
Implement feature validator utility and add tests

mlubej:


mlubej:
@zigaLuksic I had to update the functionality to use `Optional[Feature]` since we often have optional features. I think it makes sense, but let me know if you see a different way.

A simple test was added for testing a few different combinations of valid and invalid configs.

zigaLuksic:
> @zigaLuksic I had to update the functionality to use `Optional[Feature]` since we often have optional features. I think it makes sense, but let me know if you see a different way.
> 
> A simple test was added for testing a few different combinations of valid and invalid configs.

For that you can use the `optional_field_validator` utility :)

tests/test_utils/test_validators.py
```diff
@@ -267,3 +269,44 @@ class CollectionSchema(BaseSchema):
...
+    )
+
+    feature_3d: Optional[Feature]
+    _check_3d_feature = field_validator(
+        "feature_3d", restrict_types([ftype for ftype in FeatureType if ftype.ndim() == 3])
+    )
+
+
+@pytest.mark.parametrize(
+    "valid_config, invalid_config",
```
zigaLuksic, line 292: i would prefer if you split it into two tests, one for valid configurations and one for invalid ones

mlubej:
> For that you can use the optional_field_validator utility :)

Missed this one, thanks.

tests/test_utils/test_validators.py
mlubej: test split. added also a mixed usage of optional and required inputs, so that the restrictor is used in both types of field_validators.

zigaLuksic (approved):

PR #200, opened 2023-03-14
Updates needed for storing eo-patches on non AWS s3 buckets.

batic:
This PR addresses the following:

 * removes AWS specific ACLs
 * adds option to add "any FS kwargs" to storage manager
 * renames the method `is_on_aws` to `is_on_s3`, which is more appropriate
 * removes passing acl option to local fs; now adding non-appropriate kwargs to fs will fail (as it should)

With this changes, accessing eo-patches on non AWS S3 buckets is possible (e.g. creodias).

### DIFF
--- a/tests/test_core/test_storage.py
+++ b/tests/test_core/test_storage.py
@@ -114,6 +114,6 @@ def test_aws_acl(config: RawConfig):
     storage = StorageManager.from_raw_config(config)
 
     if isinstance(storage.filesystem, S3FS):
-        config_acl = config.get("filesystem_kwargs").get("acl") if config.get("filesystem_kwargs") else None
+        config_acl = config.get("filesystem_kwargs").get("acl") if "filesystem_kwargs" in config else None
         filesystem_acl = None if storage.filesystem.upload_args is None else storage.filesystem.upload_args.get("ACL")
         assert config_acl == filesystem_acl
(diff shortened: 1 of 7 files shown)


tests/test_core/test_storage.py
```diff
@@ -105,15 +105,15 @@ def test_aws_profile(aws_storage_config: RawConfig, config_profile: Optional[str
...
-        {"project_folder": ".", "aws_acl": "public-read"},
+        {"project_folder": "."},
     ],
 )
 def test_aws_acl(config: RawConfig):
     storage = StorageManager.from_raw_config(config)
 
     if isinstance(storage.filesystem, S3FS):
-        config_acl = config.get("aws_acl")
+        config_acl = config.get("filesystem_kwargs").get("acl") if config.get("filesystem_kwargs") else None
```
zigaLuksic, line 117:
```suggestion
        config_acl = config.get("filesystem_kwargs", {}).get("acl")
```

You can also do `config.get("filesystem_kwargs").get("acl") if "filesystem_kwargs" in config else None`

zigaLuksic (approved):
Love it. It's surprising how it's often simpler to do a more general version.

tests/test_core/test_storage.py
batic: Like it, using it.

PR #205, opened 2023-03-24
Add import vector pipeline

zigaLuksic:
If you'd want any additional functionality just say the word


zigaLuksic:
gotta figure out why it runs fine locally

eogrow/pipelines/import_vector.py
```diff
@@ -0,0 +1,59 @@
...
+            reproject=self.config.reproject,
+            clip=self.config.clip,
+            filesystem=self.storage.filesystem,
+        )
+        import_node = EONode(vector_import_task)
+
+        save_task = SaveTask(
+            path=self.storage.get_folder(self.config.output_folder_key),
+            filesystem=self.storage.filesystem,
+            features=[FeatureType.BBOX, FeatureType.TIMESTAMPS, self.config.output_feature],
```
devisperessutti, line 54: why do we include `FeatureType.TIMESTAMPS`, in particular when they could be empty (if I understand correctly). 

This pipeline can be run on existing eopatches, so we have the risk of overwriting the existing timestamps. For bbox is not a problem as they should be identical

PR #208, opened 2023-04-06
Toml based install

zigaLuksic:
Similar to sh-py, except here i had to explicitly exclude docs (and added tests while i was at it)
not sure why it also collects the conf.py from docs in the wheel unless excluded, didn't happen in sh-py


pyproject.toml
```diff
@@ -1,3 +1,116 @@
+[build-system]
+requires = ['hatchling']
+build-backend = 'hatchling.build'
+
+[tool.hatch.version]
+path = 'eogrow/__init__.py'
+
+
```
mlubej, line 8: one extra line here

mlubej (commented):
> Similar to sh-py, except here i had to explicitly exclude docs (and added tests while i was at it)

would it make sense to explicitly exclude it also in sh-py, even if it doesn't cause issues? did some version change or somehting?

zigaLuksic:
> > Similar to sh-py, except here i had to explicitly exclude docs (and added tests while i was at it)
> 
> would it make sense to explicitly exclude it also in sh-py, even if it doesn't cause issues? did some version change or somehting?

it's why we're doing build inspection. These things are hard to predict. It wasn't added because it seemed like the default settings are fine, but these things are a bit odd at times. We can add it, but I guess it's low priority.

pyproject.toml
zigaLuksic: done

mlubej (approved):
Tried it out, works as expected.

PR #209, opened 2023-04-06
Extend feature validator restrict_types to allow callables

mlubej:
`FeatureParser` was extended in the past to allow for callables, but this wasn't reflected in eo-grow validators.

Used the `FeatureParser` from eo-learn directly in the validator.

### DIFF
--- a/eogrow/utils/validators.py
+++ b/eogrow/utils/validators.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Tuple
 import numpy as np
 from pydantic import BaseModel, Field, root_validator, validator
 
-from eolearn.core import FeatureType
+from eolearn.core import FeatureParser, FeatureType
 from sentinelhub import DataCollection
 from sentinelhub.data_collections_bands import Band, Bands, MetaBands, Unit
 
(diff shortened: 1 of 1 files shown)


eogrow/utils/validators.py
```diff
@@ -214,12 +214,13 @@ def parse_data_collection(value: Union[str, dict, DataCollection]) -> DataCollec
...
-def restrict_types(allowed_feature_types: Iterable[FeatureType]) -> Callable[[Feature], Feature]:
+def restrict_types(
+    allowed_feature_types: Union[Iterable[FeatureType], Callable[[FeatureType], bool]]
+) -> Callable[[Feature], Feature]:
     """Validates a field representing a feature, where it restricts the possible feature types."""
 
     def validate_feature(value: Feature) -> Feature:
-        ftype, _ = value
-        assert ftype in allowed_feature_types, f"Feature type can be one of {allowed_feature_types}, got {ftype}."
+        FeatureParser(features=value, allowed_feature_types=allowed_feature_types)
```
zigaLuksic, line 223: it feel like we're relying on the current implementation of `FeatureParser` to raise the issue on `__init__`. Perhaps if we go with `parse_feature` directly?

mlubej: Agreed. Updated.

zigaLuksic (approved):

PR #213, opened 2023-04-07
Adjust compare content signature

zigaLuksic:
Compare content should be more explicit about the fact that it wants a folder.
And `run_config` can output the final folder path.

This should make it easier for more exotic test setups.

### DIFF
--- a/eogrow/utils/testing.py
+++ b/eogrow/utils/testing.py
@@ -332,42 +332,6 @@ def run_config(
         check_pipeline_logs(pipeline)
 
 
-def compare_content(
-    config_path: str,
-    stats_path: str,
-    *,
-    output_folder_key: Optional[str] = None,
-    save_new_stats: bool = False,
-) -> None:
-    """Compares the results from a pipeline run with the saved statistics
-
-    :param config_path: A path to the config file
-    :param stats_path: A path to the file containing result statistics
-    :param output_folder_key: Type of the folder containing results of the pipeline, inferred from config if missing
-    :param save_new_stats: Save new result stats and skip the comparison
-    """
-    crude_configs = collect_configs_from_path(config_path)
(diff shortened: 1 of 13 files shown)


eogrow/utils/testing.py
```diff
@@ -305,8 +306,10 @@ def run_config(
     *,
     output_folder_key: Optional[str] = None,
     reset_output_folder: bool = True,
```
mlubej, line 308: I think it would make more sense if this `reset_output_folder` would run only in the beginning, not between each config. If a test config chain is run, you're usually not interested in deleting intermediate steps. And in case you are, then you could just split the test in separate runs to achieve the same, which makes more sense anyway.

eogrow/utils/testing.py
```diff
@@ -330,41 +333,34 @@ def run_config(
...
     *,
-    output_folder_key: Optional[str] = None,
     save_new_stats: bool = False,
 ) -> None:
-    """Compares the results from a pipeline run with the saved statistics
+    """Compares the results from a pipeline run with the saved statistics. Constructed to be coupled with `run_config`
+    hence the `Optional` input.
 
-    :param config_path: A path to the config file
+    :param folder_path: A path to the folder whose contents are to be compared
```
mlubej, line 348:
```suggestion
    :param folder_path: A path to the folder with contents to be compared
```

mlubej (commented):
Looks much better and intuitive

eogrow/utils/testing.py
zigaLuksic: i actually think only the last one should be reset, since it might mess with the content tester

mlubej: Agreed to leave as is.

mlubej (approved):

PR #216, opened 2023-04-07
Add pre-commit for formatting JSON files

zigaLuksic:
also doubles as a JSON checker, and the output is very readable.

Has some additional options, but i believe the defaults are fine. Can be used for evalscripts probably


zigaLuksic:
> I was afraid that it would expand things too much, but it seems to do the opposite and the end result looks much better.

I had the same concerns, but unlike some formatters, it tries to keep the JSON human readable (to a point)

PR #218, opened 2023-04-13
Upload config files to head instead of encoding them and providing them as an encoded config string

mlubej:
Changes:
- separate cluster spin up
- use `ray rsync_up` to upload the config to the cluster head
- remove encoding/decoding functionality


mlubej:
@zigaLuksic mentioned something along the lines of trying to unify how the timestamp suffix is used throughout the repo, but I found this to be a bit messy to do, at least in this MR. The issue is that sometimes we use the timestamp format for logging (colons and spaces, e.g., `%Y-%m-%d %H:%M:%S`), sometimes we use `%Y-%m-%dT%H-%M-%S`, and sometimes we also put a `Z` at the end for Zulu time specification. The issue is that this also affects pipeline names and some test files, so it wasn't clear exactly how to unify anything.

What I think would make sense is to have `utils/time.py` file where the different formats are defined, but try to find a way to not have so much different versions of them, which would require some updates of tests.

was also thinking of a utility that returns a timestamp string from a `timestamp | None`, for a specific timestamp or the `utcnow()` timestamp, respectively.

zigaLuksic:
> @zigaLuksic mentioned something along the lines of trying to unify how the timestamp suffix is used throughout the repo, but I found this to be a bit messy to do, at least in this MR. 

I was thinking of only unifying the timestamp used for logs + config, since they are both used to generate a folder/file name that is semi-unique (as long as requests are far-between)

eogrow/cli.py
```diff
@@ -138,27 +130,38 @@ def ray(
...
 
-        configs = collect_configs_from_path(config_filename)
-        encoded_configs = encode_config_list(configs)
+        crude_configs = collect_configs_from_path(config_filename)
+        raw_configs = [interpret_config_from_dict(config) for config in crude_configs]
+        remote_path = get_cluster_config_path(config_filename)
+
+        with NamedTemporaryFile(mode="w", delete=True, suffix=".json") as local_path:
+            json.dump(raw_configs, local_path)
+            subprocess.check_call(f"ray rsync_up {cluster_yaml} {local_path.name!r} {remote_path!r}", shell=True)
```
zigaLuksic, line 153: wasn't sure whether this should be a `check_call` since we tend to use `call` elsewhere. In the [docs](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) it mentions that `run` should be used if possible.

eogrow/cli.py
```diff
@@ -138,27 +130,38 @@ def ray(
...
+        if start_cluster:
+            subprocess.check_call(f"ray up -y {cluster_yaml}", shell=True)
+
         if stop_cluster and (use_screen or use_tmux):
             raise NotImplementedError("It is not clear how to combine stop flag with either screen or tmux flag")
 
-        configs = collect_configs_from_path(config_filename)
-        encoded_configs = encode_config_list(configs)
+        crude_configs = collect_configs_from_path(config_filename)
+        raw_configs = [interpret_config_from_dict(config) for config in crude_configs]
```
zigaLuksic, line 148: these two can maybe be a single line.

eogrow/cli.py
```diff
@@ -138,27 +130,38 @@ def ray(
...
+        """Command for running an eo-grow pipeline on a remote Ray cluster of AWS EC2 instances. The provided config is
+        fully constructed and uploaded to the cluster head in the `~/.synced_configs/` directory, where it is then
+        executed. A custom suffix is added to distinguish runs which use the same config multiple times.
 
         \b
         Example:
             eogrow-ray cluster.yaml config_files/config.json
         """
+        if start_cluster:
+            subprocess.check_call(f"ray up -y {cluster_yaml}", shell=True)
```
zigaLuksic, line 142: what happens if the cluster is up? will this reset it? it's been a while since i worked with ray

mlubej: Hmm, I though the same rules apply as before with `ray exec` but [the docs](https://docs.ray.io/en/latest/cluster/cli.html#ray-exec-doc) state that this is meant as an on-need utility, but I couldn't find the exact implementation of what it does under the hood.

`ray up` can do some restarts, which in my experience is not desired and becomes complicated if docker is used, even if `--no-restart` is provided.

perhaps it would be safer to remove this option, or run the lines above only in case the cluster is not running (checking with `ray get_head_ip`)

eogrow/cli.py
mlubej: will change to `run`, since we don't do anything with the returned object anyway

eogrow/cli.py
mlubej: done

eogrow/cli.py
mlubej: done

eogrow/cli.py
zigaLuksic: if there is an easy way to check if the cluster is online, then let's do that
otherwise let's remove --start and --stop flags IMO

mlubej: Managed to define some ray utility functions for this so that it looks OK in cli. I tested it out and it works as expected:
- managed to get both True and False for the `is_cluster_running`
- `eogrow-ray` spinning up a cluster if `--start` is provided and no cluster is running
  - if `--start` is not provided, the process fails with an appropriate error message

Check the last commit https://github.com/sentinel-hub/eo-grow/pull/218/commits/433e9d22604d1975cb37476a0d9585f5f578f640

zigaLuksic (approved):

PR #219, opened 2023-04-14
Spring cleanup of utils

zigaLuksic:
noticed we had some unused functions.
some were also simplified a bit
also updated the DEM statistics to work with develop eo-learn (shouldn't change the github CI)


mlubej (approved):
breezy

PR #222, opened 2023-04-17
Add option for users to specify a custom name for each pipeline

zigaLuksic:
So things i'm not sure about:
- should it be `custom_name`, just `name`, maybe `custom_log_name`?
- do we have this parameter as a regular pipeline parameter, or as part of the `logging` config?

### DIFF
--- a/eogrow/core/pipeline.py
+++ b/eogrow/core/pipeline.py
@@ -58,7 +58,7 @@ class Pipeline(EOGrowObject):
 
     @property
     def _pipeline_name(self) -> str:
-        return self.config.custom_name or self.__class__.__name__
+        return self.config.pipeline_name or self.__class__.__name__
 
     @classmethod
     def from_raw_config(cls: Type[Self], config: RawConfig, *args: Any, **kwargs: Any) -> Self:
--- a/eogrow/core/pipeline.py
+++ b/eogrow/core/pipeline.py
@@ -56,6 +56,10 @@ class Pipeline(EOGrowObject):
         self.area_manager: BaseAreaManager = self._load_manager(config.area, storage=self.storage)
         self.logging_manager: LoggingManager = self._load_manager(config.logging, storage=self.storage)
 
+    @property
+    def _pipeline_name(self) -> str:
+        return self.config.custom_name or self.__class__.__name__
+
     @classmethod
     def from_raw_config(cls: Type[Self], config: RawConfig, *args: Any, **kwargs: Any) -> Self:
         """Creates an object from a dictionary by constructing a validated config and use it to create the object."""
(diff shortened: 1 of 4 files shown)


mlubej:
> should it be custom_name, just name, maybe custom_log_name?

what about `pipeline_name`? so that we have:

```json
{
  ...
  "pipeline": "timeseries.pipelines.download.SURSDownloadPipeline",
  "pipeline_name": "landsat-5-signals-download"
  ...
 }
```

> do we have this parameter as a regular pipeline parameter, or as part of the logging config?

I was thinking along the lines of regular optional parameter, but not sure about the pros/cons of having it a part of the logging config. The above snippet is more intuitive to me, as it is next to the pipeline itself.

eogrow/core/pipeline.py
```diff
@@ -222,7 +226,7 @@ def run(self) -> None:
                 pipeline_timestamp=timestamp,
             )
 
-            LOGGER.info("Running %s", self.__class__.__name__)
+            LOGGER.info("Running %s", self._pipeline_name)
```
mlubej, line 229: I would still keep the original line, and rather print the custom name additionally in case of providing specifics. I think the class name does bring value already on its own.

perhaps the `custom_name` is not a suitable name of the parameter, since the user would in such cases want to provide extra information about the pipeline? 

As mentioned in the example internally, in case of the following:

```
[
  {"**DEM": "${config_path}/01_download_DEM.json"},
  {"**RABA": "${config_path}/02_rasterize_RABA.json"},
  {"**mowing": "${config_path}/02_rasterize_mowing_parcels_2019.json"},
  {"**GERK": "${config_path}/02_rasterize_GERK_SEGMENT_2019.json"},
  {"**travniki": "${config_path}/02_rasterize_travniki.json"},
  {"**convert": "${config_path}/03_convert.json"}
]
```

the pipelines from the second to second-last are all rasterization pipelines, and I would like to know they are rasterization pipelines, but additionally also would like to have the option to specify the details of what kind of rasterization I'm doing, so that it's easier to find the logs and interpret the logs

we can also have a discussion if that would be easier

eogrow/core/schemas.py
```diff
@@ -27,6 +27,7 @@ class PipelineSchema(BaseSchema):
     """Base schema of the Pipeline class."""
 
     pipeline: Optional[ImportPath] = Field(description="Import path to an implementation of Pipeline class.")
+    custom_name: Optional[str] = Field(description="Custom name for the pipeline for easier identification in logs.")
```
mlubej, line 30:
```suggestion
    custom_name: Optional[str] = Field(description="Custom pipeline name for easier identification in logs.")
```

eogrow/core/pipeline.py
mlubej: discusse in private, resolving

PR #223, opened 2023-04-18
Switch to field validators from root validators

zigaLuksic:
Root validators make sense and are great, BUT they leave the docs messy.
Now the user needs to be careful of the order of parameters, but the docs are much nicer.

Before, each field was marked as having a validator (since there was a root validator)
image

Now only one of the relevant fields is marked. The validator also has a custom name (cause i felt like being extra today).
image

### DIFF
--- a/eogrow/pipelines/batch_to_eopatch.py
+++ b/eogrow/pipelines/batch_to_eopatch.py
@@ -2,7 +2,7 @@
 from typing import Any, List, Optional
 
 import numpy as np
-from pydantic import Field, root_validator
+from pydantic import Field, validator
 
 from eolearn.core import (
     CreateEOPatchTask,
@@ -47,10 +47,6 @@ class BatchToEOPatchPipeline(Pipeline):
         input_folder_key: str = Field(description="Storage manager key pointing to the path with Batch results")
         output_folder_key: str = Field(description="Storage manager key pointing to where the EOPatches are saved")
 
-        mapping: List[FeatureMappingSchema] = Field(
-            description="A list of mapping from batch files into EOPatch features."
-        )
-
         userdata_feature_name: Optional[str] = Field(
             description="A name of META_INFO feature in which userdata.json would be stored."
         )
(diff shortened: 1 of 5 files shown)


eogrow/pipelines/batch_to_eopatch.py
```diff
@@ -61,19 +57,24 @@ class Schema(Pipeline.Schema):
             ),
             example="\"[info['date'] for info in json.loads(userdata['metadata'])]\"",
         )
+
+        mapping: List[FeatureMappingSchema] = Field(
+            description="A list of mapping from batch files into EOPatch features."
+        )
+
+        @validator("mapping")
+        def check_something_is_converted(cls, value: list, values: RawSchemaDict) -> list:
```
mlubej, line 66: I find this name confusing and not really sure what the purpose of the validator is, could you suggest a better name?

eogrow/utils/validators.py
```diff
@@ -51,34 +51,44 @@ def optional_validator(value, values, config, field):  # type: ignore[no-untyped
...
 
-    def ensure_exclusion(cls: type, values: RawSchemaDict) -> RawSchemaDict:
-        is_param1_defined = values.get(param1) is None
-        is_param2_defined = values.get(param2) is None
+    Make sure that the definition of `second_param` comes after `first_param˙ (in line-order).
+    """
+
+    def ensure_exclusion(cls: type, value: Optional[Any], values: RawSchemaDict) -> Optional[Any]:
+        is_param1_defined = values.get(first_param) is None
+        is_param2_defined = value is None
```
mlubej, line 62: Doesn't change anything down the line I think, but these should probably be negated, or the name changed

eogrow/utils/validators.py
```diff
@@ -51,34 +51,44 @@ def optional_validator(value, values, config, field):  # type: ignore[no-untyped
...
         assert (
             is_param1_defined != is_param2_defined
-        ), f"Exactly one of parameters `{param1}` and `{param2}` has to be specified."
+        ), f"Exactly one of parameters `{first_param}` and `{second_param}` has to be specified."
+
+        return value
 
-        return values
+    ensure_exclusion.__name__ = f"cannot_be_used_with_{first_param}"  # used for docbuilding purposes
```
mlubej, line 69: in most cases we have a single underline used in the validators, should it be used here and in L89 as well?

eogrow/pipelines/batch_to_eopatch.py
zigaLuksic: I am almost feeling like this should just be removed. Perhaps `check_nonempty_input`?

eogrow/utils/validators.py
zigaLuksic: we never really decided on whether the underscore should be used. I used the underscore so that it is more clear what the difference between a validator and a field is, but all of these utility functions result in non-underscored validators in docs. From the docs perspective the underscores shouldn't be added.

eogrow/utils/validators.py
zigaLuksic: adjusted names, it is indeed wrong to have it like it is. Probably a mistake due to simplifications

eogrow/pipelines/batch_to_eopatch.py
mlubej: it definitely sounds better, yes

zigaLuksic: done

mlubej (approved):

PR #224, opened 2023-04-20
Refactor rasterization tests

mlubej:
Changes:
- rasterization tests were converted from config chains to single configs
- all the magic for running custom processes before the rasterization pipeline were handled in the following ways:
  - preparation configs were defined, or
  - custom columns were added to the vector files

Tried to do it in the way that the stats change minimally, at least for other pipelines.

I still think it could be done better, e.g. the resolution and the shape tests are very similar, but a lot of tests build on top of each other, sometimes resetting the output folder and sometimes not, so it's hard to find a proper way how to replace this.

Let me know if the current situation is good enough.


tests/test_pipelines/test_import_vector.py
```diff
@@ -7,7 +7,7 @@
 
 @pytest.mark.chain
 @pytest.mark.parametrize("experiment_name", ["import_vector", "import_vector_temporal"])
-def test_import_tiff_pipeline(config_and_stats_paths, experiment_name):
```
zigaLuksic, line 10: :scream_cat:

tests/test_project/input-data/test_area_lulc.geojson
zigaLuksic, line 1: should we perhaps shorten this file a bit?
4k lines for tests is a bit much

zigaLuksic (approved):
great, much simpler, a step in the right direction

tests/test_project/input-data/test_area_lulc.geojson
mlubej: I agree that the file might be too large, but the issue is that a lot of tests depend on it.

I tried reducing this and it affects 16 tests which would need updated stats.

If the MR is otherwise good to go, then I guess it's fine to squeeze in a last commit which is basically just reducing the geojson and updating the stats.

zigaLuksic: we can also modify that in a separate MR. just to keep stuff clean

PR #227, opened 2023-04-21
Validate folder key presence in storage structure

mlubej:
Changes:
- added validator for checking folder key presence
- added test

Comments:
- folder keys can be optional (sometimes defaults to "input_data")
- "input_data" key was manually added to be OK (usually not present in the structure, since it's a default)
  - not sure if there is a better way. One could use `PRESET_FOLDERS` from the storage manager, but was afraid of creating circular imports
- storage structure presence is not checked in the validator, since this should already be covered

If this addition makes sense, I can then then open a new MR where this validator is applied where it makes sense

### DIFF
--- a/eogrow/utils/validators.py
+++ b/eogrow/utils/validators.py
@@ -94,12 +94,10 @@ def ensure_defined_together(first_param: str, second_param: str, **kwargs: Any) 
 def ensure_storage_key_presence(key: str, **kwargs: Any) -> classmethod:
     """A field validator that makes sure that the specified storage key is present in the storage structure."""
 
-    def validate_storage_key(cls: type, key: str, values: RawSchemaDict) -> str:
-        storage = values.get("storage")
-        assert storage is not None, "Storage schema not found in the config!"
-
-        storage_keys = list(storage.structure.keys()) + ["input_data"]
-        assert key in storage_keys, f"Couldn't find storage key {key!r} in the storage structure!"
+    def validate_storage_key(cls: type, key: Optional[str], values: RawSchemaDict) -> Optional[str]:
+        if key is not None:
+            storage_keys = list(values["storage"].structure.keys()) + ["input_data"]
+            assert key in storage_keys, f"Couldn't find storage key {key!r} in the storage structure!"
         return key
 
     return field_validator(key, validate_storage_key, **kwargs)
--- a/eogrow/utils/validators.py
+++ b/eogrow/utils/validators.py
@@ -91,6 +91,20 @@ def ensure_defined_together(first_param: str, second_param: str, **kwargs: Any) 
     return field_validator(second_param, ensure_both, **kwargs)
 
 
+def ensure_storage_key_presence(key: str, **kwargs: Any) -> classmethod:
+    """A field validator that makes sure that the specified storage key is present in the storage structure."""
+
+    def validate_storage_key(cls: type, key: str, values: RawSchemaDict) -> str:
+        storage = values.get("storage")
+        assert storage is not None, "Storage schema not found in the config!"
+
+        storage_keys = list(storage.structure.keys()) + ["input_data"]
+        assert key in storage_keys, f"Couldn't find storage key {key!r} in the storage structure!"
+        return key
+
+    return field_validator(key, validate_storage_key, **kwargs)
+
+
 def parse_time_period(value: Tuple[str, str]) -> TimePeriod:
     """Allows parsing of preset options of shape `[preset_kind, year]` but that requires `pre` validation"""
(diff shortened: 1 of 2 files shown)


tests/test_utils/test_validators.py
```diff
@@ -137,6 +139,21 @@ class DummySchema2(BaseSchema):
         DummySchema2()
 
 
+@pytest.mark.parametrize(
+    "folder_key,raises_error", [("data", False), ("input_data", False), (None, False), ("i_do_not_exist", True)]
+)
+def test_ensure_storage_key_presence(config, folder_key, raises_error):
```
zigaLuksic, line 145: what is `config`?

eogrow/utils/validators.py
```diff
@@ -91,6 +91,18 @@ def ensure_both(cls: type, value: Optional[Any], values: RawSchemaDict) -> Optio
     return field_validator(second_param, ensure_both, **kwargs)
 
 
+def ensure_storage_key_presence(key: str, **kwargs: Any) -> classmethod:
+    """A field validator that makes sure that the specified storage key is present in the storage structure."""
+
+    def validate_storage_key(cls: type, key: Optional[str], values: RawSchemaDict) -> Optional[str]:
```
zigaLuksic, line 97: a possible ouchie (that likely cant happen, but i want a test to show that it doesnt) is that you have a  `input_folder: str` but then this validator let's a `None` through...
Very likely pydantic doublechecks the final type, but I want guarantees in form of tests

eogrow/utils/validators.py
```diff
@@ -91,6 +91,18 @@ def ensure_both(cls: type, value: Optional[Any], values: RawSchemaDict) -> Optio
...
+        if key is not None:
+            storage_keys = list(values["storage"].structure.keys()) + ["input_data"]
```
zigaLuksic, line 99: what about logs and cache?

you can also drop the `keys` btw. I'd even do something like `assert key in storage.structure or key in predefined_keys` (with apropriate variables)

zigaLuksic (commented):
My main concern here is the field order in advanced config joins. But it sounds like it's worth the hassle. Lets try and go with this.

tests/test_utils/test_validators.py
mlubej: It's a fixture for a global config, containing all managers information. here I use it to bring in the storage structure information, which better simulates what happens in actual pipelines

eogrow/utils/validators.py
mlubej: wasn't sure if those are expected as folder keys. do you think there is a way to get the preset folders of the storage manager so that I don't cause a circular import?? or is the hardcoding here acceptable?

zigaLuksic: i'd say hardcoding is ok, cause i really dont know where else to put this. We might want to reevaluate the structure

tests/test_utils/test_validators.py
mlubej: yes, exactly. would you prefer the structure is more explicitly set? it could load the config to get all the necessary pipeline parameters, but then explicitly reset the storage structure

eogrow/utils/validators.py
mlubej: This would be an issue if we would use the validator with `pre=True`, right? Here the parameter value `None` should fail even before the validation is reached.

having a test for this feels like testing the pydantic model itself, not the validator.

This is just my thinking, if you think it's still better to have an extra test for this, I'll gladly add it.

Otherwise I also tried this and I get what one would expect:

```
E   pydantic.error_wrappers.ValidationError: 1 validation error for DummySchema
E   folder_key
E     none is not an allowed value (type=type_error.none.not_allowed)
```

tests/test_utils/test_validators.py
mlubej: in the end I just explicitly updated the `config` to 

```
config["storage"]["structure"] = {"i_exist": "foobar"}
```

because adding a `dict` param to the dummy schema doesn't work, since the validator works with a storagemanager schema object. I could convert it to a dict, but that already seemed like cooking.

eogrow/utils/validators.py
mlubej: added

eogrow/utils/validators.py
mlubej: updated

tests/test_utils/test_validators.py
zigaLuksic: ah right, should be fine now

zigaLuksic (approved):

PR #228, opened 2023-04-21
Reduce test geojson size and update affected stats

mlubej:
Changes:
- reduce test vectors with intersection + zonal merging
- rerun tests and update changed stats


zigaLuksic:
what about the crops file?

PR #245, opened 2023-05-25
Switch to ruff

zigaLuksic:
Sadly new-style annotations clash with pydantic


zigaLuksic:
The 3.11 issue is a funky one.
In python 3.11 builtins are loaded as 'frozen' modules (see [notes](https://docs.python.org/3/whatsnew/3.11.html#faster-startup))
This means that paths of these modules are now "frozen", which is not really valid in our metamagic functions. But on the other hand, not sure if we should even worry about that, why would someone need the import-path of a builtin module in eo-grow...

zigaLuksic:
> The 3.11 issue is a funky one. In python 3.11 builtins are loaded as 'frozen' modules (see [notes](https://docs.python.org/3/whatsnew/3.11.html#faster-startup)) This means that paths of these modules are now "frozen", which is not really valid in our metamagic functions. But on the other hand, not sure if we should even worry about that, why would someone need the import-path of a builtin module in eo-grow...

@mlubej  please check the [fix](https://github.com/sentinel-hub/eo-grow/pull/245/commits/a47c54b23f575822ec252eadbe54b22c11e5cbfe)

eogrow/utils/meta.py
```diff
@@ -73,8 +72,10 @@ def get_os_import_path(import_path: str) -> str:
     """
     module_spec = importlib.util.find_spec(import_path)
     if module_spec is not None and module_spec.origin is not None:
+        if module_spec.origin == "frozen":  # python 3.11 optimizations preload certain core modules
+            raise ValueError(f"Given import path {import_path!r} belongs to a 'frozen' module, no path available.")
         return module_spec.origin
-    raise ValueError(f"Given import path {import_path} not found")
+    raise ValueError(f"Given import path {import_path!r} not found")
```
mlubej, line 78: So, this does't really change anything, since we don't use import paths from builtins anyway, it just failed because of one particular test which used `os`, but the failed test itself wasn't really important, it just coincidentally used the `os` package which was affected?

mlubej (approved):

eogrow/utils/meta.py
zigaLuksic: yes, but I believe it is something that should be handled 'just in case', otherwise we'd need to add it to docs somewhere etc. Since this happens at validation time, i dont mind it failing if it detects issues.

I was also thinking that maybe we should remove this functionality... It was added as a way to make evalscripts available on the cluster, but i think we can come up with something better

PR #251, opened 2023-06-26
Store cluster.yaml in logs

zigaLuksic:
We decided for a simple procedure:

1. `eogrow-ray` transfers `cluster.yaml` onto the head instance (similar to configs). This does not need to be named in a 'execution-specific-way' because if you switch to a different cluster configuration you'll (almost surely) also switch the head node
2. Each pipeline checks if there is a `cluster.yaml` in the expected location, if there is, it copies it over.

Tested it out on a simple case to make sure it works

### DIFF
--- a/eogrow/cli.py
+++ b/eogrow/cli.py
@@ -9,6 +9,7 @@ from typing import Optional, Tuple
 import click
 
 from .core.config import collect_configs_from_path, interpret_config_from_dict
+from .core.logging import CLUSTER_FILE_LOCATION
 from .core.schemas import build_schema_template
 from .pipelines.testing import TestPipeline
 from .utils.general import jsonify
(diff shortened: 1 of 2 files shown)


veseln:
Just one question. Does this work if you spin a cluster, attach to it and then run a pipeline?  Not a problem if it doesn't, just want to understand. 

```
ray up cluster.yaml
ray attach cluster.yaml 
eogrow something
```

zigaLuksic:
> Just one question. Does this work if you spin a cluster, attach to it and then run a pipeline? Not a problem if it doesn't, just want to understand.
> 
> ```
> ray up cluster.yaml
> ray attach cluster.yaml 
> eogrow something
> ```

If you do a manual `ray rsync_up cluster.yaml cluster.yaml "~/.synced_configs/cluster.yaml"` then it will (or if you first run a pipeline with `eogrow-ray`which copies it)

veseln:
Thanks, looks good.

veseln (approved):

PR #254, opened 2023-07-18
Updates for Pydantic V2

mlubej:
This was quite a journey. Had to rely a lot on https://docs.pydantic.dev/latest/migration/ which came quite handy

Main changes:
- Optional fields require an explicit `None` in the `Field`
- `validator` changed to `field_validator`, along with a different function fingerprint, so I had to update everything
  - this clashed with our own version of the `field_validator`, so I decided to switch the two names, so ours is now just `validator` and `optional_validator`
- had to update the process for schema building, because some attributes (like `type_` and `outer_type_`) are not available anymore
- I updated some deprecated methods
- random changes across the repo due to the V2 updates

I suggest going through the commits, I tried to keep them grouped in scope.


mlubej:
GDAL installation seems to fail for some reason...

mlubej:
This shouldn't be merged until `ray` fixes issues with `pydantic>=2.0`. Ray has just release a version 2.6, where they fixed the pydantic version to `<2.0`, which probably means this will not be fixed any time soon, also due to ray depending on other packages (such as `fastapi`) which also need to fix the issue.

Then there is also a question of compatibility with `pydantic<2.0`.

Perhaps it would be better to just downgrade pydantic for now, also because it's casing issues in other packages.

mlubej:
Also, I'm not sure why the ray tests pass here ... seems sometimes the ray cluster is successfully initialized, but not always? dunno

zigaLuksic:
will wait for ray to move to pydantic 2, this will come in handy in the future

PR #255, opened 2023-07-25
Restric pydantic to version 1 and apply fixes for GDAL installation

zigaLuksic:


.github/workflows/ci_action.yml
```diff
@@ -84,11 +84,11 @@ jobs:
         run: |
           sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable -y
           sudo apt-get update
-          sudo apt-get install -y build-essential gdal-bin libgdal-dev graphviz proj-bin gcc libproj-dev libspatialindex-dev
+          sudo apt-get install -y build-essential gdal-bin libgdal-dev graphviz proj-bin gcc libproj-dev libspatialindex-dev  libboost-all-dev
           export CPLUS_INCLUDE_PATH=/usr/include/gdal
           export C_INCLUDE_PATH=/usr/include/gdal
           pip install -e .[DEV,ML]
-          pip install gdal==$(gdal-config --version | awk -F'[.]' '{print $1"."$2}')
+          pip install gdal==$(gdal-config --version)
```
mlubej, line 91: Just some context here, the installation was failing when I tried it, then by trial and error found that adding the `libbost-all-dev` fixed the issue, but not sure what exactly was missing.

as for the version, I think it's better to install the exact version of GDAL which is on the system. Before this, the patch system was ignored and I think I noticed older versions being installed.

mlubej (approved):

PR #256, opened 2023-07-28
Shorten content-comparator for tests, make it safe from changes in 1.5.0

zigaLuksic:


mlubej (approved):
Seems clearer to me 👍

PR #258, opened 2023-08-01
Improve test data generation

zigaLuksic:
In this MR I only introduce a new data generation pipeline (a slightly polished old version) to keep the diffs small-ish
After the changes have been accepted, we can open a new MR where we transition to the new pipeline.


eogrow/tasks/testing.py
```diff
@@ -66,24 +79,78 @@ def execute(self, eopatch: EOPatch, seed: int | None = None) -> EOPatch:
         return eopatch
 
 
-class DummyTimestampFeatureTask(EOTask):
+class GenerateRasterFeatureTask(EOTask):
```
mlubej, line 82: Just to mention that we already have some similar code for the same functionality in https://github.com/sentinel-hub/eo-learn/blob/develop/core/eolearn/core/utils/testing.py#L47. Perhaps this was missed, or perhaps the functionality is different. Would it make sense to make the two parts of code more aligned?

eogrow/pipelines/testing.py
```diff
@@ -83,6 +89,36 @@ class TimestampFeatureSchema(FeatureSchema):
...
+    std: float = Field(1, description="Standard deviation of the normal distribution.")
+
+
+class RasterFeatureGenerationSchema(BaseSchema):
+    feature: Feature = Field(description="Feature to be created.")
+    shape: Tuple[int, ...] = Field(description="Shape of the feature")
+    dtype: np.dtype = Field(description="The output dtype of the feature")
+    _parse_dtype = field_validator("dtype", parse_dtype, pre=True)
+    distribution: Union[UniformDistributionSchema, NormalDistributionSchema] = Field(
+        description="Choice of distribution for generating values.", discriminator="kind"
```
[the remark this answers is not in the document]
mlubej: I checked:

```json
{
  "pipeline": "eogrow.pipelines.testing.GenerateDataPipeline",
  "pipeline_name": "<< Optional[str] >>",
  "storage": {
    "manager": "<< Optional[str] >>"
  },
  "area": {
    "manager": "<< Optional[str] >>"
  },
  "logging": {
    "manager": "<< Optional[str] >>"
  },
  "workers": "<< 1 : int >>",
  "use_ray": "<< 'auto' : Union[Literal['auto'], bool] >>",
  "test_subset": "<< Optional[List[Union[int, str]]] >>",
  "skip_existing": "<< bool >>",
  "output_folder_key": "<< str >>",
  "seed": "<< int >>",
  "features": {
    "<< type >>": "List[RasterFeatureGenerationSchema]",
    "<< nested schema >>": "<class 'eogrow.pipelines.testing.RasterFeatureGenerationSchema'>",
    "<< sub-template >>": {
      "feature": "<< Tuple[FeatureType, str] >>",
      "shape": "<< Tuple[int, ...] >>",
      "dtype": "<< dtype >>",
      "distribution": "<< Union[UniformDistributionSchema, NormalDistributionSchema] >>"
    }
  },
  "timestamps": {
    "time_period": "<< Tuple[date, date] >>",
    "num_timestamps": "<< int >>",
    "same_for_all": "<< True : bool >>"
  },
  "meta_info": "<< Optional[dict] >>"
}
```

should this be going more in depth or is it fine?

eogrow/tasks/testing.py
zigaLuksic: the functionality in eo-learn focuses on generating dummy eopatches, but here we focus on generating features. Some key differences:
- allows per-feature configuration
- allows different distributions
- more precise configuration

in eo-learn we have no need of that, so I wouldn't join them too much

eogrow/pipelines/testing.py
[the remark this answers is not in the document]
zigaLuksic: we might want to check if the template generator knows how to handle Unions properly. but i wouldn't lose sleep over that, it's just templates

eogrow/pipelines/testing.py
```diff
@@ -165,3 +201,81 @@ def get_execution_arguments(self, workflow: EOWorkflow, patch_list: PatchList) -
...
+            return NormalDistribution(distribution_config.mean, distribution_config.std)
+        return UniformDistribution(distribution_config.min_value, distribution_config.max_value)
+
+    def get_execution_arguments(self, workflow: EOWorkflow, patch_list: PatchList) -> ExecKwargs:
+        """Extends the basic method for adding execution arguments by adding seed arguments a sampling task"""
+        exec_args = super().get_execution_arguments(workflow, patch_list)
+
+        rng = np.random.default_rng(seed=self.config.seed)
+        per_node_seeds = {node: rng.integers(low=0, high=2**32) for node in workflow.get_nodes()}
+        same_timestamps = self.config.timestamps and self.config.timestamps.same_for_all
```
zigaLuksic, line 267: `same_for_all` is used here.

tests/tasks/test_testing.py
```diff
@@ -71,3 +79,82 @@ def test_dummy_timestamp_feature_task(dummy_eopatch: EOPatch):
...
+    eopatch = task.execute(dummy_eopatch.copy(), seed=seed)
+
+    assert feature in eopatch
+    assert len(eopatch.get_features()) == 3
+
+    data: np.ndarray = eopatch[feature]
+    assert data.shape == shape
+    assert data.dtype == dtype
+    if not is_discrete_type(dtype):
+        assert shapiro(data)[1] > 0.05, "Shapiro-Wilks test rejects that the data is normally distributed."
```
mlubej, line 153: if I understand correctly, the Shapiro test is a test for normality of the data. Perhaps you could use the [Kolmogorov-Smirnov](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html) test, which doesn't depend on the distribution, and then you could use it in both cases, for the normal one as well as the uniform one, and the test could be simplified.

for example:

```python
from scipy import norm, uniform, kstest

data_norm = norm.rvs(size=30*60, loc=1, scale=0.5)
data_uni = uniform.rvs(size=30*60, loc=0, scale=2)

# good
kstest(data_uni, stats.uniform.cdf, args=(0, 2))
Out[84]: KstestResult(statistic=0.01889668569539571, pvalue=0.5350380083686563, statistic_location=1.5489044825019025, statistic_sign=-1)

# bad
kstest(data_norm, stats.uniform.cdf, args=(0, 2))
Out[85]: KstestResult(statistic=0.10682627834303782, pvalue=2.4249614533423762e-18, statistic_location=1.4252363322028132, statistic_sign=1)

# bad
kstest(data_uni, stats.norm.cdf, args=(1, 0.5))
Out[86]: KstestResult(statistic=0.10891666539212586, pvalue=4.735882711535022e-19, statistic_location=1.5100065744155635, statistic_sign=-1)

# good
kstest(data_norm, stats.norm.cdf, args=(1, 0.5))
Out[87]: KstestResult(statistic=0.025362671109045776, pvalue=0.1938881590422188, statistic_location=1.3109111847328712, statistic_sign=1)
```

tests/tasks/test_testing.py
```diff
@@ -71,3 +79,82 @@ def test_dummy_timestamp_feature_task(dummy_eopatch: EOPatch):
...
+    [
+        (FeatureType.DATA, (10, 20, 21, 5), np.float32, -3, -1.5),
+        (FeatureType.DATA_TIMELESS, (83, 69, 1), int, -2, 5),
+        (FeatureType.MASK, (10, 4, 16, 7), np.int8, -3, 7),
+        (FeatureType.LABEL, (10, 271), bool, 0, 1),
+    ],
+)
+@pytest.mark.parametrize("seed", [1, 22, 9182])
+def test_generate_raster_feature_task_uniform(dummy_eopatch, feature_type, shape, dtype, min_value, max_value, seed):
+    dummy_eopatch.timestamps = ["2011-08-12"] * 10
```
mlubej, line 95: would it be better to use the info from shape here?

```suggestion
    dummy_eopatch.timestamps = ["2011-08-12"] * shape[0]
```

mlubej: Ah, not in all the cases..

tests/tasks/test_testing.py
zigaLuksic: switched to K-S for uniform and chi-square for the discrete uniform

zigaLuksic: will try to join the tests as well

zigaLuksic: joined, switched the normal distribution to k-s as well, so that we have more guarantee that the mean and std are correct.

tests/tasks/test_testing.py
```diff
@@ -71,3 +79,50 @@ def test_dummy_timestamp_feature_task(dummy_eopatch: EOPatch):
...
+    assert data.dtype == dtype
+
+    if is_discrete_type(data.dtype):
+        if isinstance(distribution, UniformDistribution):
+            _, p_value = chisquare(np.unique(data, return_counts=True)[1])
+            assert p_value > 0.05
+        else:
+            # "normally distributed" integers are just a hack, so we test what we can
+            assert np.mean(data) == pytest.approx(distribution.mean, abs=0.2)
+            assert np.std(data) == pytest.approx(distribution.std, abs=0.2)
```
mlubej, line 118: should the error used here be relative and not absolute? Just in case the tests change or something

mlubej: Additionally, this could in principle also be tested with the chi-square, but due to the complicated nature of the rounded numbers from the normal distribution let's not go there..

zigaLuksic: yeah, i'd rather not calculate the correct distribution for this by hand :see_no_evil:

zigaLuksic: for the relative vs. absolute, i'd argue absolute, because there is no reason the `mean` should be more accurate if it's closer to 0

mlubej: I guess we can keep both at `abs`, but we replace the hard-coded values with the expected standard error of the mean and of the st. dev, based on the number of samples.

If I'm not mistaken, the equations are:

```math
SE(\bar{x}) = \frac{\sigma_{x}}{\sqrt{n}}
```

```math
SE(\sigma_{x}) = \frac{\sigma_{x}}{\sqrt{2(n-1)}}
```

These are 95% confidence intervals and can be used for the `abs` value in the approximations.

Taken from https://www.tqmp.org/RegularArticles/vol10-2/p107/p107.pdf

can you compare these values to the 0.2 based on the data from the test?

PR #266, opened 2023-08-23
Improve statistics generating

zigaLuksic:
This should make them a bit less volatile and easier to understand.


mlubej (approved):
Pretty straight-forward.

I still have minor concerns for cases which are highly imbalanced, where it would be more likely that the 10% sample fails to pick some underrepresented classes, but that's more a problem of the test, which shouldn't test such an input IMO.

PR #267, opened 2023-08-23
Apply config comprehension to dictionary keys

zigaLuksic:


eogrow/core/config.py
```diff
@@ -164,7 +164,7 @@ def _sub_variable(match: re.Match, variable_mapping: dict[str, str]) -> str:
 def _recursive_apply_to_strings(config: object, function: Callable) -> object:
     """Recursively applies a function on all string values (and not keys) of a nested config object"""
     if isinstance(config, dict):
-        return {key: _recursive_apply_to_strings(value, function) for key, value in config.items()}
+        return {function(key): _recursive_apply_to_strings(value, function) for key, value in config.items()}
```
mlubej, line 167: the `_recursive_apply_to_strings` function is also used with other functions, where perhaps it might not make sense to apply it on keys. Should perhaps this function be extended with a boolean parameter whether to apply it also on keys?

zigaLuksic: it is used for:
- resolving config paths
- resolving import paths
- resolving variables

I don't see a usecase for the first two, but i also dont see a reason to not allow it. Was thinking about the same thing tho

mlubej: It was just a thought, probably it's fine, as you also pointed out. I guess also that the inclusion of keys in the recursive approach is more intuitive than the exclusion.

mlubej (approved):

PR #276, opened 2023-09-29
Refactor testing utils

zigaLuksic:
Before we do any work on the testing utils I decided to do some refactoring.

I removed the unnecessary class (there was nothing class-like about it).
This will also make it easier to add custom tailoring to `compare_content` etc.


mlubej (approved):
So many fewer indents <3

PR #277, opened 2023-10-04
Add a flag to use zarr arrays

zigaLuksic:
I am unsure about the parameter name. I was also considering the `use_zarr` flag, but i'm worried that this flag would be used in some pipelines to also use zarr optimizations.

Open to suggestions.

The tests do not include this, since Zarr is not a requirement of eo-grow. I turned this flag on in the global config and the whole thing worked automagically.

### DIFF
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -41,6 +41,7 @@ class StorageManager(EOGrowObject):
         geopandas_backend: Literal["fiona", "pyogrio"] = Field(
             "fiona", description="Which backend is used for IO operations when using geopandas."
         )
+        save_with_zarr: bool = Field(False, description="Whether to save EOPatches with the Zarr backend.")
 
         class Config(ManagerSchema.Config):
             case_sensitive = True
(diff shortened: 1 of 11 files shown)


mlubej:
> I was also considering the use_zarr flag, but i'm worried that this flag would be used in some pipelines to also use zarr optimizations.

Can you elaborate on this use case? Do you mean if someone would add this flag to their pipeline schema, even if it can already be switched on with the storage parameter?

zigaLuksic:
> > I was also considering the use_zarr flag, but i'm worried that this flag would be used in some pipelines to also use zarr optimizations.
> 
> Can you elaborate on this use case? Do you mean if someone would add this flag to their pipeline schema, even if it can already be switched on with the storage parameter?

For instance in FD where we do predictions, one can also turn on predicting on chunks. But i guess you can just name that 'chunk-predictions' or whatever. Probably me overthinking it, so it's your job to tell me to stop overthinking it :)

mlubej:
> For instance in FD where we do predictions, one can also turn on predicting on chunks. But i guess you can just name that 'chunk-predictions' or whatever. Probably me overthinking it, so it's your job to tell me to stop overthinking it :)

yes, I vote to rename to `use_zarr` :3 specific purposes should be marked with specific parameter names

PR #280, opened 2023-10-05
Switch to pydantic 2

zigaLuksic:
Started separately from what @mlubej already did, but used it as a cheat-sheet (came in handy with aliases)

The hardest part was adding all those Nones.
The template generating is a bit hacked, borrowed some ideas from @mlubej but also adjusted it to my own liking a bit. It should now work with multiple kinds of nested schemas (at a tiny cost of an extra indentation)

Thanks for paving the way @mlubej 

(i will also check docs)

### DIFF
--- a/eogrow/core/schemas.py
+++ b/eogrow/core/schemas.py
@@ -24,15 +24,15 @@ BaseSchema = EOGrowObject.Schema
 class ManagerSchema(BaseSchema):
     """A basic schema for managers, to be used as a parent class for defining manager schemas"""
 
-    manager: Optional[ImportPath] = Field(description="An import path to this specific manager.")
+    manager: Optional[ImportPath] = Field(None, description="An import path to this specific manager.")
 
 
 class PipelineSchema(BaseSchema):
     """Base schema of the Pipeline class."""
 
-    pipeline: Optional[ImportPath] = Field(description="Import path to an implementation of Pipeline class.")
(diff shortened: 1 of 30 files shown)


eogrow/core/schemas.py
```diff
@@ -24,25 +25,25 @@
 class ManagerSchema(BaseSchema):
     """A basic schema for managers, to be used as a parent class for defining manager schemas"""
 
-    manager: Optional[ImportPath] = Field(description="An import path to this specific manager.")
+    manager: Optional[ImportPath] = Field(None, description="An import path to this specific manager.")
```
mlubej, line 28: are the new type hints not applied to field types in this case? or are we not using that here?

mlubej (approved):
Didn't dive so deeply in the template generation, but I checked a few templates and they look good.

eogrow/core/schemas.py
zigaLuksic: you mean the new-style syntax of `str | None`? won't work until 3.10, 'future annotations' backports it but only during typechecking.

zigaLuksic: I guess we could increase it to 6, since it's usually between 6 and 7 digits, but I went with 5 to allow some more numerical differences

zigaLuksic:
decided to wait longer (yet again) since we need a hack to deal with docs and due to ray misbehaving occasionally.

PR #282, opened 2023-10-09
Adjust precision of test statistics

zigaLuksic:
There are quite a few design choices being done here, so point out everything that bothers you.


mlubej (approved):
IMO the stats aren't any harder to look at now as before, so that's a plus. I think it makes sense that the amount of decimal points goes down with the larger values of floats, since precision of larger float values is also lower, as was already privately discussed.

My only concern is that the 5/10 significant digits sometimes wouldn't be enough or something, especially with larger values, but that's probably just a false feeling, since the relative error limit is always the same here, if I understand correctly

zigaLuksic:
> My only concern is that the 5/10 significant digits sometimes wouldn't be enough or something, especially with larger values, but that's probably just a false feeling, since the relative error limit is always the same here, if I understand correctly

Yeah, 5 digits seems really low, but that's float32...

zigaLuksic:
we could increase it to 6 and 12, which are the edges of significant digit precision, but it could be that stuff would differ due to chip differences already.

mlubej:
> we could increase it to 6 and 12, which are the edges of significant digit precision, but it could be that stuff would differ due to chip differences already.

Mah, I'd leave it, I don't think we need to be _that_ precise for tests.

PR #283, opened 2023-10-10
1.6.1 polish

zigaLuksic:
DId you know the docs for predictions and training were not even building :upside_down_face: funky stuff


zigaLuksic:
tested that docs build with this setup

eogrow/utils/testing.py
```diff
@@ -168,10 +168,14 @@ def _calculate_tiff_stats(tiff_filename: str, config: StatCalcConfig) -> JsonDic
 def _calculate_vector_stats(gdf: gpd.GeoDataFrame, config: StatCalcConfig) -> JsonDict:
     """Calculates statistics over a vector GeoDataFrame"""
 
-    def _rounder(x: float, y: float) -> tuple[float, float]:
+    def _rounder(coords: tuple[float, float]) -> tuple[float, float]:
+        x, y = coords
         return _prepare_value(x, np.float64), _prepare_value(y, np.float64)
```
mlubej, line 173: if this `_rounder` function is meant only for vector stats anyway, how about extending it to work also on `Point` objects, since intuitively it doesn't really make a difference

```suggestion
    def _rounder(coords: tuple[float, float] | Point) -> tuple[float, float]:
        x, y = (coords.x, coords.y) if isinstance(coords, Point) else coords
        return _prepare_value(x, np.float64), _prepare_value(y, np.float64)
```

since this then makes it look better down the line (will post suggestions)

eogrow/utils/testing.py
```diff
@@ -168,10 +168,14 @@ def _calculate_tiff_stats(tiff_filename: str, config: StatCalcConfig) -> JsonDic
...
 
-    gdf.geometry = gdf.geometry.apply(lambda geometry: shapely.ops.transform(_rounder, geometry))
+    def _get_coords(geom: Polygon | MultiPolygon | Any) -> list[tuple[float, float]] | None:
```
mlubej, line 175: I would rename this function to `_get_coords_sample` to more explicitly show that only a subset of points is taken.

eogrow/utils/testing.py
```diff
@@ -183,9 +187,10 @@ def _rounder(x: float, y: float) -> tuple[float, float]:
 
     if len(gdf):
         subsample: gpd.GeoDataFrame = gdf.sample(min(len(gdf), config.num_random_values), random_state=42)
-        subsample["centroid"] = subsample.centroid.apply(lambda point: _rounder(*point.coords[0]))
+        subsample["centroid"] = subsample.centroid.apply(lambda point: _rounder(point.coords[0]))
```
mlubej, line 190: With the change to `_rounder` proposed above, this could then be 

```suggestion
        subsample["centroid"] = subsample.centroid.apply(_rounder)
```

PR #285, opened 2023-10-11
Add flag for testing logs in content comparison

zigaLuksic:
Always great to find another issue the same day you release a patch version

### DIFF
--- a/eogrow/utils/testing.py
+++ b/eogrow/utils/testing.py
@@ -265,6 +265,8 @@ def run_config(
     :param config_path: A path to the config file
     :param output_folder_key: Type of the folder containing results of the pipeline, inferred from config if None
     :param reset_output_folder: Delete the content of the results folder before running the pipeline
+    :param check_logs: If pipeline logs should be checked after the run completes. If EOWorkflows were used, the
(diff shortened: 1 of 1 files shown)


eogrow/utils/testing.py
```diff
@@ -256,6 +256,7 @@ def run_config(
     *,
     output_folder_key: str | None = None,
     reset_output_folder: bool = True,
+    check_logs: bool = True,
```
mlubej, line 259: the parameter description in the docs is missing

mlubej:
Can you elaborate a bit on what was the issue? Is it bad that we always checked logs (since it was hard-coded)? Did this cause issues sometimes?

zigaLuksic:
> Can you elaborate a bit on what was the issue? Is it bad that we always checked logs (since it was hard-coded)? Did this cause issues sometimes?

If a pipeline does not use EOWorkflows (like the one for merging vectors) then it produces an empty `finished.json` file. Checking pipeline logs fails because 'no executions finished'.

mlubej:
I just started working on the other MR now and noticed this, thanks!

mlubej (approved):

PR #288, opened 2023-10-25
Make patch_list generating faster

zigaLuksic:
This took a bit of experimenting but the end result is that in a grid of 2 x 100 000 patches, the time to generate the patch list goes from 18s to 1.75s :smile_cat: 

The idea was to stay in geopandas as long as possible. Some other things I tried:
- Mapping directly on `grid.geometry` kinda works, but since `BBox` is not a valid geometry, an error gets raised (and silenced), but in the process of raising this error a deprecation warning for the BBox string representation pops up :see_no_evil:. I mean we could silence that, but it also only gets us to 2.7s
- Then i tried applying a helper function directly onto the `grid.bounds`. This got me to 2.1s, but the code was long and the helper function couldn't be a lambda (does not capture the crs), so it was an additional definition which had to be wrapped by `partial` etc. So it was actually more code than the current solution.
- Tried to preprocess as much as possible (get stuff to a series of tuples) in a pandas-compatible way. I then remembered that `partial` also works on classes. So this was clean and fast :smile_cat: 

I did some minor tests and I don't think we can get it to work much faster. Extracting the boundary tuples from the grid takes a bit under 1s, and just creating 200 000 bboxes takes a bit under 1s. So this seems to be about as good as we can get without a brand new idea.

### DIFF
--- a/eogrow/core/area/base.py
+++ b/eogrow/core/area/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import logging
 from abc import ABCMeta, abstractmethod
+from functools import partial
 from typing import Literal, Optional
 
 import fiona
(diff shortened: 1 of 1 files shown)


mlubej (approved):
This is pretty cool, good job. A very significant improvement with not much code change

PR #289, opened 2023-10-26
Pipelines can request ray worker types

zigaLuksic:


eogrow/core/schemas.py
```diff
@@ -47,6 +47,13 @@ class PipelineSchema(BaseSchema):
     workers: int = Field(
         1, description="Number of workers for parallel execution of workflows. Parameter does not affect ray clusters."
     )
+    worker_type: Optional[str] = Field(
```
mlubej, line 50: I'm just thinking about how this generalizes, or if there is a need to be generalized:
- will we have tasks where we can specify multiple suitable worker types?
- will we have tasks that will _require_ multiple custom resources? e.g. if we ever need to parallelize training over multiple folds on an instance with multiple GPUs, once could create a custom resource of 1 GPU and 1 CPU

eogrow/core/pipeline.py
```diff
@@ -171,8 +171,11 @@ def run_execution(
         executor_class: type[EOExecutor]
 
         execution_kind = self._init_processing()
+        extra_kwargs = {}
         if execution_kind is ProcessingType.RAY:
             executor_class = RayExecutor
+            if self.config.worker_type is not None:
+                extra_kwargs = {"ray_remote_kwargs": {"resources": {self.config.worker_type: 0.001}}}
```
mlubej, line 178: Do we need to check how this affects scheduling? i.e. when the cluster is empty (head only) ray starts creating new nodes based on the amount of tasks in the queue, and by default this takes into account how many CPUs there are on the machine (or how many were specified to be used by the resources tag). My concern here would be that this means that the scheduler would think "OK, I can only create 1 new node, since that covers the 1000 cases", while in fact this would then represent a bottleneck, because the new node only has e.g. 8 CPUs

eogrow/core/schemas.py
mlubej, line 50: should this name be more specific to ray, since it's used exclusively for ray, or do we wish to keep it general in case we ever extend the functionality to other cluster platforms?

zigaLuksic: this doesn't generalize. we chose to keep it simple at this point.
If we ever need more elaborate resource requests, then this field should just be `ray_remote_kwargs` and nothing else. But then you'd have to say `"ray_remote_kwargs" = {"resources": {"BigRamWorker": 1}}` instead of `"worker_type" = "BigRamWorker"} and it's also less intuitive what we're trying to achieve here.

For the first bullet point, workers can have multiple tags. So if a pipeline can be run on multiple workers, just create a custom tag for it and add it to those workers.

Not sure what you mean with the second one

eogrow/core/pipeline.py
zigaLuksic: Every remote worker still requests 1 CPU by default. If you are worried about this we can add this explicitly in eo-learn.

eogrow/core/schemas.py
zigaLuksic: i can add a `ray` prefix if you'd like
it's still short enough

eogrow/core/pipeline.py
zigaLuksic: So just to make clear, `options` only override the parameters that were given. Since we do not pass new requirements for CPUs they stay at 1.
image

eogrow/core/schemas.py
mlubej: another example pops to mid. Say we have a big ram worker `r5.2xlarge` (8 CPUs, 64 GB RAM) and one pipeline requires 25 GB per task, while another requires 50 GB per task.

do we:
- create two different worker types in the YAML (`BigRamWorkerA` (with 32 GB instance) and `BigRamWorkerB` (with 64 GB instance)), where worker node has the amount of CPU resources set to 1 in both cases, or
- could we say for pipeline A that we need `"ray_resources" = {"BigRamWorker": 0.5, "CPU": 4}` and for pipeline B `"ray_resources" = {"BigRamWorker": 1,  "CPU": 8}`

In the second suggestion, (if I didn't mess something up), we could be running both pipelines efficiently and on the same worker (since there would be 2 processes for pipeline A running on the worker at the same time)

mlubej: but, this requires you know how many CPUs are on the worker.

zigaLuksic: there is also an option to request memory as a resource, so that would probably fit better :)

as i said, there was a discussion on which path to choose, and we said we'd go with the 'worker_type' for now. It's simple to change it to full resource management later, but you can also re-open that discussion

mlubej: due to some issue with notifications, I missed the discussion, so i will respect the decisions made there. Going with the first option here is completely valid as well

mlubej: I feel it might be misleading otherwise.. idk. if we ever switch to resources, it may be easier to go from `ray_worker_type` to `ray_resources`

zigaLuksic: It's really something that we can easily change in the future (it's just how the parameters are passed from the config to the executor) so i wouldn't lose sleep over it.

zigaLuksic: done

eogrow/core/schemas.py
```diff
@@ -47,6 +47,13 @@ class PipelineSchema(BaseSchema):
     workers: int = Field(
         1, description="Number of workers for parallel execution of workflows. Parameter does not affect ray clusters."
     )
+    ray_worker_type: Optional[str] = Field(
+        description=(
+            "Restricts execution of parallelized parts only to worker instances of the requested type. To declare the"
+            " type of a worker, tag by specifying (in the `cluster.yaml`) that it has a resource with a name matching"
+            " the type and set it to 1."
```
mlubej, line 54: Might need formatting, so adding this as a single sentence:

"Restricts execution of parallelized tasks only to `ray` worker instances of the requested type. The worker section of the `cluster.yaml` file should specify the custom resource with a matching name and the value set to 1."

mlubej (approved):

eogrow/core/schemas.py
```diff
@@ -47,6 +47,13 @@ class PipelineSchema(BaseSchema):
     workers: int = Field(
         1, description="Number of workers for parallel execution of workflows. Parameter does not affect ray clusters."
     )
+    ray_worker_type: Optional[str] = Field(
+        description=(
+            "Restricts execution of parallelized tasks only to ray worker instances of the requested type. The worker"
+            " section of the cluster.yaml file should specify the custom resource with a matching name and the value"
+            " set to 1."
```
mlubej, line 54: seems like the back ticks fell out from `ray` and `cluster.yaml`. Please add them in if you think it makes sense to have them.

mlubej (approved):

PR #291, opened 2023-11-06
Patch for new eo learn version

zigaLuksic:
We can either merge it now or wait for the new version (the only reason to wait is the VSI not being enabled but that would only matter if someone would run stuff on develop this week)


mlubej (approved):
Approving, but I'd rather wait, unless you also tested locally. I'm afraid I can't rely on my tests anymore

PR #299, opened 2023-11-10
Make logging a tiny bit more customizable

zigaLuksic:
This let's us use the methods in other parts as well.

Tried to keep it simple.


mlubej (approved):
Makes more sense that the logger object is passed around, yes

PR #300, opened 2023-11-10
Accept ray as the one and only saviour of our souls

zigaLuksic:
# IMPORTANT NOTE:
Do not review this on friday!

### Less important notes:
- rather smooth in general
- the sample-merging pipeline has some duplicate code now, but the pipeline itself is not that important. I'd even say that we should change how the data is loaded anyway, eoexecution seems overkill
- should we add a more informative error if the cluster is not running?


zigaLuksic:
the tests seem to fail due to a 3hour old release of dataclasses json which sounds like a monday problem :upside_down_face:

.github/workflows/ci_action.yml
```diff
@@ -90,6 +90,9 @@ jobs:
           pip install -e .[DEV,ML]
           pip install gdal==$(gdal-config --version)
 
+      - name: Set up local cluster
+        run: ray start --head
```
mlubej, line 94: You mentioned issues with initializing ray programmatically from python with `ray.init()`. Is that still relevant and would that not be the preferred way of running things on a local ray cluster by default?

Or is this exact logic implemented, and you're just making sure the cluster is set up separately for github actions?

docs/source/common-configuration-patterns.md
```diff
@@ -9,8 +9,6 @@ Invoking `eogrow-template "eogrow.pipelines.zipmap.ZipMapPipeline" "zipmap.json"
 {
   "pipeline": "eogrow.pipelines.zipmap.ZipMapPipeline",
   "pipeline_name": "<< Optional[str] >>",
-  "workers": "<< 1 : int >>",
```
mlubej, line 12: This would now be controlled via the ray remote kwargs, right?

mlubej: it might be cumbersome in the future to provide cpu workers via ray resources. we might need to think about some more convenient ways to set number of workers or something.

eogrow/pipelines/merge_samples.py
```diff
@@ -43,7 +44,6 @@ class Schema(Pipeline.Schema):
         )
         suffix: str = Field("", description="String to append to array filenames")
         workers: int = Field(1, description="Number of threads used to load data from EOPatches in parallel.")
```
mlubej, line 46: should we rename this parameter to be more in line with the number of threads for loading?  `threads`/`loading_threads`/`threads_per_worker`?

eogrow/core/pipeline.py
```diff
@@ -198,18 +183,13 @@ def run_execution(
...
 
         successful = [execution_names[idx] for idx in executor.get_successful_executions()]
         failed = [execution_names[idx] for idx in executor.get_failed_executions()]
-        LOGGER.info(
-            "%s finished with %d / %d success rate",
-            executor_class.__name__,
-            len(successful),
-            len(successful) + len(failed),
-        )
+        LOGGER.info("EOExecutor inished with %d / %d success rate", len(successful), len(successful) + len(failed))
```
mlubej, line 192:
```suggestion
        LOGGER.info("EOExecutor finished with %d / %d success rate", len(successful), len(successful) + len(failed))
```

.github/workflows/ci_action.yml
mlubej: > IMO the cluster should be set up in whatever way the user wants,

The user wants it to be a black box 😈 

In all seriousness, it just feels that doing `ray start --head` every time you want to run something is a bit cumbersome for the users. Doing it automatically under the hood is more in line with the current user experience, but I agree that having the option for the user to start the cluster manually should also be supported.

[According to ray](https://docs.ray.io/en/latest/ray-core/starting-ray.html#starting-ray-on-a-single-machine) it seems a completely legit way of initializing the cluster.

Perhaps what I'm suggesting is not the best solution, since we have to try first how this change will behave anyway, but in that case maybe we're missing some instructions what the user should do in order to start the cluster, if the cluster can't be found

docs/source/common-configuration-patterns.md
mlubej: In the past we had the option to set the amount of CPU workers via the `workers` parameter. This parameter is not deprecated, but controlling CPU workers can still be done via the `ray_remote_kwargs`. I just wanted to point out that it might be a bit unintuitive to provide CPU workers that way in case of unskilled users. What do you think?

.github/workflows/ci_action.yml
zigaLuksic: from what we experimented with back in the day using `ray.init()` did not connect you to the remote cluster, it just created a local one. We had to add the `address="auto"` which then blocks if it cannot find a running cluster.

And judging by the documentation that you linked, nothing has changed.

Since eo-grow is for grown-ups (and very talented younglings) i think they can manage a `ray start --head` once a day when running things locally :P

docs/source/common-configuration-patterns.md
zigaLuksic: i don't think performance tweaking is meant for unskilled users. They can also set up the local ray cluster with restricted CPUs.

.github/workflows/ci_action.yml
zigaLuksic: i re-read the whole docs, and it might be that it works now. I would still prefer for the user to spin up the cluster themselves, but you can create a new MR and test out the possible scenarios and if all work, i am willing to merge it.

eogrow/core/pipeline.py
```diff
@@ -169,26 +163,17 @@ def run_execution(
...
+        LOGGER.info("Starting processing for %d EOPatches", len(execution_kwargs))
 
         # Unpacking manually to ensure order matches
         list_of_kwargs, execution_names = [], []
         for exec_name, exec_kwargs in execution_kwargs.items():
             list_of_kwargs.append(exec_kwargs)
             execution_names.append(exec_name)
 
-        executor = executor_class(
+        executor = RayExecutor(
```
mlubej, line 176: One last concern I have is with debugging, currently we could run the pipeline locally and just debug using the ipdb/vscode debugger, but this might cause problems now since it's wrapped in ray.

There are some mentions that for local ray clusters the behavior should be fine, or that one can use the [ray debugger](https://docs.ray.io/en/latest/ray-observability/user-guides/debug-apps/ray-debugging.html) but I didn't manage to stop inside an eogrow pipeline test, which is worrying.

.github/workflows/ci_action.yml
mlubej, line 94: this is now also necessary to do before running tests and from a researcher/user point-of-view I would expect that the testing ray cluster would be set up in the conftest or something.

I'm now starting to get worried that I got prematurely excited and helped push us in this direction 😬 let's hope these are just temporary obstacles which need a small paradigm shift

zigaLuksic: it's simple, you expect too much :P

the debugging part is IMO very crucial, and i'll try to resolve it. Without it this change should not yet happen (or we should at least include a 'debug' mode which runs it with an EOExecutor in  a single thread)

eogrow/pipelines/merge_samples.py
```diff
@@ -59,9 +59,36 @@ def run_procedure(self) -> tuple[list[str], list[str]]:
...
+            filesystem=self.storage.filesystem,
+            logs_filter=EOExecutionFilter(ignore_packages=self.logging_manager.config.eoexecution_ignore_packages),
+            logs_handler_factory=EOExecutionHandler,
+            raise_on_temporal_mismatch=self.config.raise_on_temporal_mismatch,
+        )
+        execution_results = executor.run(multiprocess=True, workers=self.config.num_threads)
+
+        successful = [execution_names[idx] for idx in executor.get_successful_executions()]
+        failed = [execution_names[idx] for idx in executor.get_failed_executions()]
+        LOGGER.info("EOExecutor inished with %d / %d success rate", len(successful), len(successful) + len(failed))
```
mlubej, line 85:
```suggestion
        LOGGER.info("EOExecutor finished with %d / %d success rate", len(successful), len(successful) + len(failed))
```

eogrow/core/schemas.py
```diff
@@ -79,6 +69,9 @@ class PipelineSchema(BaseSchema):
     raise_on_temporal_mismatch: bool = Field(
         False, description="Treat `TemporalDimensionWarning` as an exception during EOExecution."
     )
+    debug: bool = Field(
+        False, description="Run pipeline with less parallelization to enable coupling with the debugger."
```
mlubej, line 73:
```suggestion
        False, description="Run pipeline without parallelization to enable debugging."
```

zigaLuksic: well....
it's not true :see_no_evil: it only disable parallelization of the eoexecutor...

mlubej: Updated, hope it's better

eogrow/core/pipeline.py
mlubej: the debugging now works for me after the recent changes. I think it's good enough for now.

.github/workflows/ci_action.yml
mlubej: As discussed, there are still things to be discussed. I'm fine with this MR being merged and we can make updates on top of it, or we wait with the merging until we figure out the logging and storage manager stuff, in case that affects the pipelines

eogrow/core/schemas.py
zigaLuksic: not entirely accurate but whatever, accepted

mlubej (approved):
Would it make sense to mention this in our team so that others test it out as well?

either before or after merging

zigaLuksic:
> Would it make sense to mention this in our team so that others test it out as well?
> 
> either before or after merging

i think it would make a lovely surprise :P

what kind of testing do you have in mind?

PR #303, opened 2023-11-15
Run pipeline on remote

zigaLuksic:
When using CLI commands the pipeline initialization is done on a worker process.

The process can request ray remote resources via (multiple) `-r` options, e.g. `-r "num_cpus:3" -r "resources.my_resource:1"`.
With this the main pipeline process can be requested to run on a worker machine.

If `debug: True` is detected in a pipeline config the process will not be started in a worker process (making it easier to use `ipdb`).

~Currently using `eogrow config.json` will spawn a local cluster if it does not exist yet. This might be a win for researchers, but after some testing i got 72 idle ray processes so I'm inclined to put brakes on that.~
The cluster has to be started manually even when running locally. I get lots of zombie processes if i let python do it.

Any names are up to debate, I think there is a lot of repetition between these and the pipeline parameters.

The one downside i see is that all stdout logs now have a worker-process prefix (file logs are fine). Maybe we could/should get rid of that? takes up valuable screen space.

### DIFF
--- a/eogrow/cli.py
+++ b/eogrow/cli.py
@@ -1,10 +1,10 @@
 """Implements the command line interface for `eo-grow`."""
 
-import ast
 import json
 import os
 import re
 import subprocess
+from collections import defaultdict
 from tempfile import NamedTemporaryFile
 from typing import Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


zigaLuksic:
a downside of this approach is that in chain configs one cannot specify main-process resources per-pipeline.

However this can probably be alleviated by simply having worker instances be big and the worker instance with the main-process can also process eopatches on the side (one can reserve 16GB memory for the main process and appropriately less worker processes will be run on that worker)

zigaLuksic:
in the end we decided to go with config chain improvements

PR #304, opened 2023-11-17
Better chain configs

zigaLuksic:
Alternative to https://github.com/sentinel-hub/eo-grow/pull/303

The change is a bit larger, but now the pipeline chain config can store remote execution parameters

### DIFF
--- a/eogrow/cli.py
+++ b/eogrow/cli.py
@@ -1,16 +1,18 @@
 """Implements the command line interface for `eo-grow`."""
 
+from __future__ import annotations
+
 import json
 import os
 import re
 import subprocess
 from tempfile import NamedTemporaryFile
-from typing import Optional, Tuple
 
 import click
 
-from .core.config import collect_configs_from_path, interpret_config_from_dict
+from .core.config import CrudeConfig, RawConfig, collect_configs_from_path, interpret_config_from_dict
 from .core.logging import CLUSTER_FILE_LOCATION_ON_HEAD
+from .core.pipeline_chain import validate_chain
 from .core.schemas import build_schema_template
 from .pipelines.testing import TestPipeline
 from .utils.general import jsonify
@@ -42,7 +44,7 @@ test_patches_option = click.option(
 @click.argument("config_path", type=click.Path())
 @variables_option
 @test_patches_option
-def run_pipeline(config_path: str, cli_variables: Tuple[str, ...], test_patches: Tuple[int, ...]) -> None:
+def run_pipeline(config_path: str, cli_variables: tuple[str, ...], test_patches: tuple[int, ...]) -> None:
     """Execute eo-grow pipeline using CLI.
 
     \b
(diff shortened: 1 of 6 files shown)


zigaLuksic:
I was thinking of using `pipeline_process_resources` and `worker_resources` instead of the `ray_remote_kwargs` etc.

Sure we'll have `pipeline_process_resources: {"resources: {"A": 1}}` but in general it will be more intuitive?

zigaLuksic:
forgot to mention, if this does not get merged, we need to port the fix for `eogrow-ray` + CLI variables over

eogrow/cli.py
```diff
@@ -99,11 +101,9 @@ def run_pipeline_on_cluster(
     if start_cluster:
         start_cluster_if_needed(cluster_yaml)
 
-    raw_configs = [interpret_config_from_dict(config) for config in collect_configs_from_path(config_path)]
     remote_path = generate_cluster_config_path(config_path)
-
     with NamedTemporaryFile(mode="w", delete=True, suffix=".json") as local_path:
-        json.dump(raw_configs, local_path)
+        json.dump(collect_configs_from_path(config_path), local_path)
```
mlubej, line 106: this now dumps the crude configs, right? I guess that's more in line with the configs we have on disk anyway, but is there a specific reason for the change?

zigaLuksic: previously the variables were not instantiated correctly. If you did `eogrow-ray -v "x:1"` it wouldn't apply the variable, because it interpreted the config without the CLI variables

The `eogrow-ray` command is just used to execute the appropriate `eogrow` command on the cluster, and that command expects a crude config (since **it** does the instantiation).

eogrow/utils/pipeline_chain.py
```diff
@@ -0,0 +1,41 @@
...
+import ray
+from pydantic import Field, ValidationError
+
+from ..core.config import RawConfig
+from ..core.schemas import BaseSchema
+from .meta import collect_schema, load_pipeline_class
+
+
+class PipelineRunSchema(BaseSchema):
+    pipeline_config: dict
```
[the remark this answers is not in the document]
zigaLuksic: done

zigaLuksic:
I'll try to squeeze in a test as well, never a bad thing

zigaLuksic:
> I'll try to squeeze in a test as well, never a bad thing

the tests i added are rather simplistic, but it's still better than nothing IMO

zigaLuksic:
just remembered that the docs need to be updated as well, doing that

docs/source/config-language.md
```diff
@@ -26,19 +26,28 @@ Additional notes:
 
 ### Pipeline chains
 
-A typical configuration is a dictionary with pipeline parameters. However, it can also be a list of dictionaries. In this case each dictionary must contain parameters of a single pipeline. The order of dictionaries defines the consecutive order in which pipelines will be run. Example:
+A typical configuration is a dictionary with pipeline parameters. However, it can also be a list of pipeline-execution dictionaries that specify:
+- `pipeline_config`: a configuration for a single pipeline,
+- `pipeline_resources`: a dictionary that is passed to `ray.remote` to configure which resources the main pipeline process will request from the cluster (see [here](https://docs.ray.io/en/latest/ray-core/api/doc/ray.remote_function.RemoteFunction.options.html) for options)
```
mlubej, line 31: should probably mention what the default is if this is not provided

mlubej (approved):

PR #309, opened 2023-11-22
Fix pipeline-chain validation

zigaLuksic:
As per usual, one finds the bug just after release :upside_down_face: 

The problem is that variables are handled with `interpret_config_from_dict` which pops them directly from the config. In this case the config is `{"pipeline_config": ...}` and thus does not have variables (which are one level lower). The interpretation then fails because it finds variables that it does not have defined (because it didnt manage to pop the correct variable dict). Tried to get around that.

Do you think this is important enough to release 1.7.1 tomorrow?

### DIFF
--- a/eogrow/cli.py
+++ b/eogrow/cli.py
@@ -207,7 +207,11 @@ def validate_config(config_path: str) -> None:
         pipeline_config = _prepare_config(config, {}, ())
         collect_schema(load_pipeline_class(pipeline_config)).parse_obj(pipeline_config)
     else:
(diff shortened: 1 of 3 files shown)


mlubej (approved):
Good catch, even if it's after the release. Are the tests missing this use case? I thought we had some chain validation checks before, so I hadn't expected this.

I think this one might be important enough to make a minor release

zigaLuksic:
> Are the tests missing this use case? I thought we had some chain validation checks before, so I hadn't expected this.

The tests work on the helper function, but this was a CLI issue. I am tempted to add a CLI test for this, and i probably will

mlubej (approved):

PR #315, opened 2023-12-08
Better rounding in tests

zigaLuksic:
Also includes the sh-py fix.

### DIFF
--- a/eogrow/core/storage.py
+++ b/eogrow/core/storage.py
@@ -7,7 +7,6 @@ from typing import Any, ClassVar, Dict, Literal, Optional
 import fs
 from pydantic import BaseSettings, Field
 
-import sentinelhub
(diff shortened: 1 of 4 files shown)


tests/pipelines/test_byoc.py
```diff
@@ -124,12 +126,41 @@ def test_temporal_byoc(config_and_stats_paths, preparation_config, config, confi
     assert check_request.method == "GET"
 
     timestamps = [
```
mlubej, line 128: why did these timestamps change?

zigaLuksic: because the `data_2019` was thrown out when we reduced download tests, so it's now using another downloaded data set

zigaLuksic: it's not in the pipeline chain and thus never got tested in CI. Kinda concerning that we never ran it locally :upside_down_face:

PR #330, opened 2024-02-21
Support regular parquet in test statistics

zigaLuksic:
**WARNING:** if you review this before Monday I will draw an unflattering doodle of you. Do not play games with me!

Not sure if we should have a special case where we still try to load the parquet as a geoparquet. Seems kinda hard to do in general.

### DIFF
--- a/eogrow/utils/testing.py
+++ b/eogrow/utils/testing.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import json
 import os
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from functools import partial
 from typing import Any, Iterable, Optional, cast
 
@@ -92,8 +92,8 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
         elif content_path.endswith((".geojson", ".gpkg")):
             stats[content] = _calculate_vector_stats(gpd.read_file(content_path, engine="pyogrio"), config)
         elif content_path.endswith(".parquet"):
-            config = StatCalcConfig(**config.dict(), num_random_values=0)  # due to non-determinism in parquet row order
-            stats[content] = _get_parquet_stats(content_path, config)
+            # don't sample random rows due to non-determinism in parquet row order
+            stats[content] = _get_parquet_stats(content_path, replace(config, num_random_values=0))
         else:
             stats[content] = None
 
--- a/eogrow/utils/testing.py
+++ b/eogrow/utils/testing.py
@@ -213,7 +213,7 @@ def _calculate_parquet_stats(data: pd.DataFrame, config: StatCalcConfig) -> Json
     stats = {"columns": list(data), "row_count": len(data)}
 
     if len(data):
-        subsample: pd.GeoDataFrame = data.sample(min(len(data), config.num_random_values), random_state=42)
+        subsample: pd.DataFrame = data.sample(min(len(data), config.num_random_values), random_state=42)
         for col in subsample.select_dtypes(include="number").columns.values:
             subsample[col] = subsample[col].apply(partial(_prepare_value, dtype=subsample[col].dtype))
 
--- a/eogrow/utils/testing.py
+++ b/eogrow/utils/testing.py
@@ -19,7 +19,7 @@ import rasterio
 from deepdiff import DeepDiff
 from fs.base import FS
 from fs.osfs import OSFS
-from shapely import MultiPolygon, Point, Polygon, wkb, wkt
+from shapely import MultiPolygon, Point, Polygon
 
 from eolearn.core import EOPatch, FeatureType
 from eolearn.core.eodata_io import get_filesystem_data_info
@@ -93,22 +93,14 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
         elif content_path.endswith(".parquet"):
             try:
                 data = gpd.read_parquet(content_path)
+                stats[content] = _calculate_vector_stats(data, config)
             except Exception:
-                data = _load_as_geoparquet(content_path)
-            stats[content] = _calculate_vector_stats(data, config)
+                data = pd.read_parquet(content_path)
+                stats[content] = _calculate_parquet_stats(data, config)
         else:
             stats[content] = None
 
     return stats
-
-
-def _load_as_geoparquet(path: str) -> gpd.GeoDataFrame:
-    data = pd.read_parquet(path)
-    if isinstance(data.geometry.iloc[0], str):
-        data.geometry = data.geometry.apply(wkt.loads)
-    elif isinstance(data.geometry.iloc[0], bytes):
-        data.geometry = data.geometry.apply(wkb.loads)
-    return gpd.GeoDataFrame(data, geometry="geometry", crs=data.utm_crs.iloc[0])
 
 
 def _calculate_eopatch_stats(eopatch: EOPatch, config: StatCalcConfig) -> JsonDict:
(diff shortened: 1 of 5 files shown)


zigaLuksic:
i tested this in FB and the geometry column is problematic, not jsonifiable...

eogrow/utils/testing.py
```diff
@@ -217,6 +209,19 @@ def _get_coords_sample(geom: Polygon | MultiPolygon | Any) -> list[tuple[float,
     return stats
 
 
+def _calculate_parquet_stats(data: pd.DataFrame, config: StatCalcConfig) -> JsonDict:
+    stats = {"columns": list(data), "row_count": len(data)}
+
+    if len(data):
+        subsample: gpd.GeoDataFrame = data.sample(min(len(data), config.num_random_values), random_state=42)
```
mlubej, line 216: why do you cast this as `gpd.GeoDataFrame` here? Couldn't we leave it as `pd.DataFrame`? Does it perhaps auto convert the wkb/wkt into geometries?

zigaLuksic: because pandas/geopandas dont expose their types and mypy didnt catch that the copy-pasted annotation was wrong :)

zigaLuksic: but it was just that, an annotation, didnt change the fact that it was a pandas dataframe

eogrow/utils/testing.py
```diff
@@ -213,7 +213,7 @@ def _calculate_parquet_stats(data: pd.DataFrame, config: StatCalcConfig) -> Json
     stats = {"columns": list(data), "row_count": len(data)}
 
     if len(data):
-        subsample: gpd.GeoDataFrame = data.sample(min(len(data), config.num_random_values), random_state=42)
+        subsample: pd.GeoDataFrame = data.sample(min(len(data), config.num_random_values), random_state=42)
```
mlubej, line 216: typo, should be `pd.DataFrame`

zigaLuksic:
@mlubej any ideas how to solve the issue of WKB not being json serializable?

zigaLuksic:
@mlubej one option i thought of was that the tester supplies the CRS. Wont work for parquets where each geometry potentially has a different CRS though...

zigaLuksic:
tested it out a bit, the wkb->wkt also works fine. Kept the user-supplied CRS so that we can have more statistics in those cases

eogrow/utils/testing.py
```diff
@@ -91,24 +92,27 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
...
+        return _calculate_vector_stats(data, config)
+    except Exception:
+        data = pd.read_parquet(content_path)
+        if config.parquet_epsg:
+            data.geometry = data.geometry.apply(wkb.loads)
+            parsed_data = gpd.GeoDataFrame(data, geometry="geometry", crs=config.parquet_epsg)
+            return _calculate_vector_stats(parsed_data, config)
+
+        if "geometry" in data:
+            data.geometry = data.geometry.apply(wkb.loads).apply(to_wkt)
```
mlubej, line 114: should we check if the geometry is in wkb, or is it OK if we assume?

zigaLuksic: by checking if the dtype is bytes?

mlubej: yeah, that makes sense

eogrow/utils/testing.py
```diff
@@ -91,24 +92,27 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
...
-        data.geometry = data.geometry.apply(wkb.loads)
-    return gpd.GeoDataFrame(data, geometry="geometry", crs=data.utm_crs.iloc[0])
+def _get_parquet_stats(content_path: str, config: StatCalcConfig) -> JsonDict:
+    try:
+        data = gpd.read_parquet(content_path)
+        return _calculate_vector_stats(data, config)
+    except Exception:
+        data = pd.read_parquet(content_path)
+        if config.parquet_epsg:
+            data.geometry = data.geometry.apply(wkb.loads)
```
mlubej, line 109: should we do similar checks as below also here, or does the logic/assumption threshold change due to the presence of the `parquet_epsg`?

zigaLuksic: in that case we could do a check whether we need to use wkb or wkt, but its starting to feel very convoluted...

mlubej: it still didn't feel completely right to me, how would you feel about the following:

```python
def _get_parquet_stats(content_path: str, config: StatCalcConfig) -> JsonDict:
    try:
        data = gpd.read_parquet(content_path)
        return _calculate_vector_stats(data, config)
    except Exception:
        data = pd.read_parquet(content_path)
        if "geometry" not in data:
            return _calculate_parquet_stats(data, config)

        loader = wkb.loads if isinstance(data.geometry.iloc[0], bytes) else wkt.loads
        data.geometry = data.geometry.apply(loader)

        if config.parquet_epsg:
            parsed_data = gpd.GeoDataFrame(data, geometry="geometry", crs=config.parquet_epsg)
            return _calculate_vector_stats(parsed_data, config)

        data.geometry = data.geometry.apply(wkt.dumps)
        return _calculate_parquet_stats(data, config)
```

mlubej:
updates (@zigaLuksic):
- added `pyogrio` as default gpd reading engine
- set the `num_random_values` to 0 for parquet stats, code block is still there and redundant
- saved also dtypes (as strings) along with columns for parquet and vector files

eogrow/utils/testing.py
```diff
@@ -89,26 +90,34 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
...
-            stats[content] = _calculate_vector_stats(gpd.read_file(content_path), config)
+            stats[content] = _calculate_vector_stats(gpd.read_file(content_path, engine="pyogrio"), config)
         elif content_path.endswith(".parquet"):
-            try:
-                data = gpd.read_parquet(content_path)
-            except Exception:
-                data = _load_as_geoparquet(content_path)
-            stats[content] = _calculate_vector_stats(data, config)
+            # don't sample random rows due to non-determinism in parquet row order
+            stats[content] = _get_parquet_stats(content_path, replace(config, num_random_values=0))
```
zigaLuksic, line 96: i was thinking of leaving this to the user

eogrow/utils/testing.py
```diff
@@ -89,26 +90,34 @@ def calculate_statistics(folder: str, config: StatCalcConfig) -> JsonDict:
...
-def _load_as_geoparquet(path: str) -> gpd.GeoDataFrame:
-    data = pd.read_parquet(path)
-    if isinstance(data.geometry.iloc[0], str):
-        data.geometry = data.geometry.apply(wkt.loads)
-    elif isinstance(data.geometry.iloc[0], bytes):
-        data.geometry = data.geometry.apply(wkb.loads)
-    return gpd.GeoDataFrame(data, geometry="geometry", crs=data.utm_crs.iloc[0])
+def _get_parquet_stats(content_path: str, config: StatCalcConfig) -> JsonDict:
+    try:
+        data = gpd.read_parquet(content_path, engine="pyogrio")
```
zigaLuksic, line 105: does pyogrio read parquets?

mlubej: thought it was `read_file`..., removed

eogrow/utils/testing.py
mlubej: failed to notice that the config is passed in `compare_content`, removed

zigaLuksic: recent improvement :)

mlubej:
Added also an `agg_stats` for dataframe-related stats and included only mean/std/min/max.

The tests for `tests/pipelines/test_import_vector.py` need to be updated, but we don't have the mechanism in place to update the stats via CI artifacts.

@zigaLuksic do you think it makes sense to add that also for eo-grow? I can prepare that in a separate MR if you agree, otherwise I can just manually update from a linux machine

PR #331, opened 2024-03-08
Store exception traceback in logs f pipeline fails

zigaLuksic:
The `finally` clause kicks in even if another exception is raised in the `except` clause

any improvement suggestions are welcome


zigaLuksic:
we could also just log the traceback, but it seemed better to put it into a separate file

mlubej (approved):
seems good, no comments

PR #337, opened 2024-03-15
Add mechanism for retrying partial batch jobs

zigaLuksic:
Cant say i like the current batch download pipeline. Tried to do a minimal adjustment, open to any suggestions. Especially on how to test this.

### DIFF
--- a/eogrow/pipelines/download_batch.py
+++ b/eogrow/pipelines/download_batch.py
@@ -96,6 +96,9 @@ class BatchDownloadPipeline(Pipeline):
                 " method of `SentinelHubBatch` during the creation process."
             ),
         )
+        num_retries_on_partial: int = Field(
+            0, ge=0, description="How many times to retry the batch job if the resulting status is PARTIAL."
+        )
 
         analysis_only: bool = Field(
             False,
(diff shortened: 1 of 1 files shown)


eogrow/pipelines/download_batch.py
```diff
@@ -162,12 +165,16 @@ def run_procedure(self) -> tuple[list[str], list[str]]:
...
-            config=self.sh_config,
-            sleep_time=self.config.monitoring_sleep_time,
-            analysis_sleep_time=self.config.monitoring_analysis_sleep_time,
-        )
+        results = self._monitor_job(batch_request)
+
+        # retry partial
+        for _ in range(self.config.num_retries_on_partial):
+            batch_request = self.batch_client.get_request(batch_request)
+            if batch_request.status != BatchRequestStatus.PARTIAL:
```
mlubej, line 173: Should this be `BatchRequestStatus.PARTIAL.value` ? I assume `batch_request.status` is a string.

zigaLuksic: in L153 a direct comparison is also made, and the type hint claims its a `BatchRequestStatus`

zigaLuksic: `status: BatchRequestStatus = field(metadata=enum_config(BatchRequestStatus))` the field uses a special constructor

eogrow/pipelines/download_batch.py
```diff
@@ -96,6 +96,9 @@ class Schema(Pipeline.Schema):
                 " method of `SentinelHubBatch` during the creation process."
             ),
         )
+        num_retries_on_partial: int = Field(
```
mlubej, line 99: Wondering if `num_retries` would be sufficient, since the description is also present. There are probably also not any other contexts where retries make sense for this pipeline

zigaLuksic: i wanted to make it clear that retries are only done when the result is `PARTIAL`
but i guess that is also in the description

mlubej (approved):

PR #350, opened 2024-06-18
Retry 3 times if batch dowload monitoring returns 404

zigaLuksic:
Went with a very simple approach to it.

### DIFF
--- a/eogrow/pipelines/download_batch.py
+++ b/eogrow/pipelines/download_batch.py
@@ -3,10 +3,12 @@
 from __future__ import annotations
 
 import logging
+import time
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


mlubej (approved):
I trust you with the types and decorator specifics, but the logic seems straight-forward enough.

Tested a simple batch download to check if anything breaks, but seems to work fine 👍

zigaLuksic:
> I trust you with the types and decorator specifics, but the logic seems straight-forward enough.
> 
> Tested a simple batch download to check if anything breaks, but seems to work fine 👍

I guess we could add a test for the decorator, but since it includes sleep times...

PR #354, opened 2024-09-27
Prepare release 1.7.12

zigaLuksic:


zigaLuksic:
numpy 2 changes the stats JUUUST a bit. So i added the requirement of `numpy>=2` to the DEV requirements, otherwise we never know which one will be used for tests.

PR #356, opened 2024-11-05
Fix bug with 404 on GET request for batch_request

mlubej:
Issue is that SH has separate read/write databases which can be out of sync. Adds a wrapper for retrying on 404 around the GET request as well

Let me know if you want to name them differently

### DIFF
--- a/eogrow/pipelines/download_batch.py
+++ b/eogrow/pipelines/download_batch.py
@@ -49,7 +49,7 @@ P = ParamSpec("P")
 def _retry_on_404(func: Callable[P, T]) -> Callable[P, T]:
     @wraps(func)
     def retrying_func(*args: P.args, **kwargs: P.kwargs) -> T:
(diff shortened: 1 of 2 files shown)


eogrow/pipelines/download_batch.py
```diff
@@ -308,6 +309,11 @@ def _trigger_user_action(self, batch_request: BatchRequest) -> BatchUserAction:
         LOGGER.info("Didn't trigger batch job because current batch request status is %s", status)
         return BatchUserAction.NONE
 
+    @_retry_on_404
```
zigaLuksic, line 312: should we tweak the wait times?

zigaLuksic (approved):

eogrow/pipelines/download_batch.py
zigaLuksic: we discussed in private and added another intermediate value

PR #365, opened 2025-03-17
Fail stats comparison if changes are present also when updating the stats

mlubej:
In the `compare_content` we have the option to set the `save_new_stats` parameter to true, which will save the new statistics. The new statistics will then be compared and the comparison should find no difference, which is the only possible conclusion, since the comparison in this case is basically the identity check and brings no value.

It makes more sense to fail the test regardless of the `save_new_stats` parameter:
- such a behavior is more in line with pre-commit/lint checks, where fixes are applied, but the job fails if changes were made
- jobs can be run only once and you can then make use of the test as well as the stats update using artifacts

### DIFF
--- a/.github/workflows/ci_action.yml
+++ b/.github/workflows/ci_action.yml
@@ -90,10 +90,8 @@ jobs:
           pip install -e .[DEV,ML]
           pip install gdal==$(gdal-config --version)
 
-      - name: Set up local cluster # we need to install async-timeout until ray 2.9.0 fixes the issue
-        run: |
-          pip install async-timeout
-          ray start --head
+      - name: Set up local cluster
+        run: ray start --head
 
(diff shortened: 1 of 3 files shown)


mlubej:
Some weird issue with the failing test where the local cluster is not being set up

zigaLuksic (approved):

mlubej:
> looks like we'll temporarily have to limit the ray version?

tried limiting it back to 2.9 but it still doesn't work, it's a weird issue.. time to drop 3.8? 😁

mlubej:
[aiohttp-cors v0.8.0](https://github.com/aio-libs/aiohttp-cors/releases) was released 2 days ago, downgrading to v0.7.0 solves the issue

PR #366, opened 2025-09-10
Remove retry on partial functionality

zigaLuksic:


zigaLuksic:
i think we're hitting issues because 3.8 is now end-of-life 
@mlubej I suggest we drop 3.8 from the test suite (and possibly package compatibility)

pyproject.toml
```diff
@@ -205,6 +202,7 @@ warn_unreachable = true
 strict_equality = true
 pretty = true
 plugins = ["pydantic.mypy"]
+files = ["eogrow"]
```
mlubej, line 205: what does this do exactly? what file is this?

zigaLuksic: this informs mypy that by default it should check files in the `eogrow` folder, so you can just write `mypy` instead of `mypy eogrow`. I've grown lazy and don't feel like typing the entire command all the time :speak_no_evil:

mlubej: I thought it might be something like that, cool

PR #374, opened 2025-09-18
Batch pipeline creates a patch list instead

mlubej:
Changes:
- instead of updating the batch grid, rather save a patch list of valid patches
- valid patches are all patches which which were successfully processed and delivered
    - the user can decide to stop processing if some tiles failed, otherwise to continue
- other pipelines need to define the patch list in order to focus only on available tiles

### DIFF
--- a/eogrow/pipelines/download_batch.py
+++ b/eogrow/pipelines/download_batch.py
@@ -295,6 +295,16 @@ class BatchDownloadPipeline(Pipeline):
         grid_path = fs.path.join(grid_folder, self.area_manager.config.grid_filename)
         save_grid(grid, grid_path, self.storage)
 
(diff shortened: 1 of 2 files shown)


eogrow/pipelines/download_batch.py
```diff
@@ -199,8 +204,7 @@ def _parse_area_name_column(cls, area: CustomGridAreaManager.Schema) -> CustomGr
                 "existing batch job. If it is not given it will create a new batch job."
             ),
         )
-        patch_list: None = None
-        input_patch_file: None = None
```
mlubej, line 203: This functionality could maybe be enforced with `area.patch_names: None = None` to ensure that no patch list is used with the batch download pipeline

